PR #36, opened 2019-04-04
Update install.md

geremyCohen:
Adding delete permissions, without them, CRD controller will not be able to delete App Mesh resources.


jqmichael:
LGTM

jqmichael:
BTW, it might be easier to use `appmesh:*` at this point. 
The only non-touched permissions are `TagResource` and `UntagResource`
https://docs.aws.amazon.com/app-mesh/latest/APIReference/Welcome.html

nckturner:
@jqmichael @geremyCohen Let's not use `appmesh:*`.  Its better to be explicit, because `appmesh:*` will likely include more permissions in the future, and it's less visible to users what they are allowing the controller to do.

Issue #43, opened 2019-06-17
Support Logging File Path in VirtualNode configuration


stphivos:
This currently works by setting it manually from the aws console after a virtual node is created by the controller e.g.:

**Logging**
*HTTP access logs path*: `/dev/stdout`

Ideally would like to be able to specify this from the K8s CRDs.

PR #45, opened 2019-06-25
Adding logging field to virtual-node

kiranmeduri:
*Issue #, if available:*
#43 

*Description of changes:*
Adding logging field to virtual-node to match https://docs.aws.amazon.com/app-mesh/latest/APIReference/API_VirtualNodeSpec.html


pkg/aws/appmesh.go
```diff
@@ -284,6 +285,20 @@ func (c *Cloud) CreateVirtualNode(ctx context.Context, vnode *appmeshv1beta1.Vir
 		}
 	}
 
+	if vnode.Spec.Logging != nil {
+		if vnode.Spec.Logging.AccessLog != nil {
+			if vnode.Spec.Logging.AccessLog.File != nil {
```
mogren: Nit: I would prefer to join the if-statements, but maybe that's just me?

pkg/aws/appmesh.go
```diff
@@ -347,6 +362,20 @@ func (c *Cloud) UpdateVirtualNode(ctx context.Context, vnode *appmeshv1beta1.Vir
...
```
mogren: Same as above.

pkg/controller/virtualnode.go
```diff
@@ -211,6 +213,33 @@ func vnodeNeedsUpdate(desired *appmeshv1beta1.VirtualNode, target *aws.VirtualNo
...
+					}
+					if desired.Spec.Logging.AccessLog.File.Path != awssdk.StringValue(target.Data.Spec.Logging.AccessLog.File.Path) {
+						return true
+					}
+				}
+			}
+		}
+	} else if target.Data.Spec.Logging != nil {
+		return true
+	}
```
mogren: Lots of cases here, but they all add up. Could be worth adding comments, but not a must.

kiranmeduri: separated out into its own method

pkg/aws/appmesh.go
kiranmeduri: yeah joined them, in future if there are variances then we can revisit breaking them.

Issue #42, opened 2019-06-14
Fails to create a VirtualService with TCP weightedTargets


stphivos:
`TCP` routes are ignored.

**Logs**
```
...
I0614 09:11:37.120839       1 controller.go:320] Virtual Node Updated
I0614 09:11:37.121324       1 controller.go:320] Virtual Node Updated
E0614 09:11:37.127253       1 controller.go:422] error syncing 'default/myapp': error updating resource while adding finalizer virtualServiceDeletion.finalizers.appmesh.k8s.aws to virtual service myapp: VirtualService.appmesh.k8s.aws "myapp" is invalid: ... "kind":"VirtualService"}: validation failure list:
spec.routes.http.action.weightedTargets in body must be of type array: "null"
I0614 09:11:37.130980       1 controller.go:212] Pod Updated
I0614 09:11:37.130996       1 controller.go:212] Pod Updated
...
```

**Config**
```
apiVersion: appmesh.k8s.aws/v1beta1
kind: VirtualService
metadata:
  name: myapp
spec:
  meshName: mymesh
  routes:
    - name: default-route
      tcp:
        action:
          weightedTargets:
            - virtualNodeName: myapp-v1
              weight: 1
            - virtualNodeName: myapp-v2
              weight: 1
```

*Note*: This works as expected with an `http` route.

**Update**:  Validation error seems to be because `apis/appmesh/v1beta1/types.go` Route's `HttpRoute` [does not have omitempty](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/apis/appmesh/v1beta1/types.go#L111).

stefanprodan:
You virtual service is missing the virtual router, here is an working example, generated by [Flagger](https://github.com/weaveworks/flagger)

```yaml
apiVersion: appmesh.k8s.aws/v1beta1
kind: VirtualNode
metadata:
  name: podinfo-canary
  namespace: test
spec:
  listeners:
  - portMapping:
      port: 9898
      protocol: http
  meshName: global
  serviceDiscovery:
    dns:
      hostName: podinfo-canary.test
---
apiVersion: appmesh.k8s.aws/v1beta1
kind: VirtualNode
metadata:
  name: podinfo-primary
  namespace: test
spec:
  listeners:
  - portMapping:
      port: 9898
      protocol: http
  meshName: global
  serviceDiscovery:
    dns:
      hostName: podinfo-primary.test
---
[...]

PR #46, opened 2019-06-25
Added TCP route support

kiranmeduri:
*Issue #, if available:*
#42 

*Description of changes:*
Add TCP route support to match https://docs.aws.amazon.com/app-mesh/latest/APIReference/API_RouteSpec.html


examples/color.yaml.template
```diff
@@ -359,4 +363,100 @@ spec:
             - name: "COLOR"
               value: "red"
 ---
-
+apiVersion: appmesh.k8s.aws/v1beta1
+kind: VirtualNode
+metadata:
+  name: tcpecho
```
nckturner: Long term, I think we should have multiple examples that each show off a subset of related features.  This will help users determine what configuration is necessary for each feature.  I think tcpecho is a good candidate for the first stand-alone example, what do you think?

kiranmeduri: I believe examples should go to https://github.com/aws/aws-app-mesh-examples under walkthroughs.

examples/color.yaml.template
```diff
@@ -359,4 +363,100 @@ spec:
...
+spec:
+  meshName: ${MESH_NAME}
+  virtualRouter:
+    listeners:
+      - portMapping:
+          port: 2701
+          protocol: tcp
+  routes:
+    - name: tcpecho-route
+      tcp:
```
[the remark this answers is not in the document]
kiranmeduri: nope, AppMesh API will throw error.

examples/color.yaml.template
nckturner: That makes sense.  I'll create an issue. We can leave it as one example for now, then.

kiranmeduri:
Added tcp to virtual-service CRD spec. Please take a look. Thanks

PR #44, opened 2019-06-19
Added listeners to virtual-router under VirtualService CRD

kiranmeduri:
*Issue #, if available:*
#37 

*Description of changes:*
[VirtualRouter](https://docs.aws.amazon.com/app-mesh/latest/APIReference/API_VirtualRouterSpec.html) allows listener to be configured hence this  update to CRD. 

Also includes
- modified color.yaml to be a template with mesh-name, namespace and images injectable for local dev.
- updated aws sdk

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #54, opened 2019-08-09
Update readme to include virtualRouter listener under virtualService

kiranmeduri:
*Issue #, if available:*

*Description of changes:*
Related to VirtualRouter change made in #44  


stefanprodan:
I've updated Flagger to create http listeners https://github.com/weaveworks/flagger/pull/272

It would be great if the controller and CRDs would use sem ver to signal breaking changes.

kiranmeduri:
yeah something like #55. Not sure how we are doing releases in the past.

PR #56, opened 2019-08-16
Add http-header match and priority to routes

kiranmeduri:
*Issue #, if available:*

*Description of changes:*

1. Added health-check for VirtualNode.Listener
2. Add http-headers and priority support to Route.HttpRoute
3. Update aws-sdk-go to v1.23.2

Testing Done:
- Verified with local changes to aws-app-mesh-examples


kiranmeduri:
> ``.

done, thanks.

kiranmeduri:
> Can you update deploy/all.yaml?

Done

kiranmeduri:
> Currently that file is referenced by the [our install guide](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/docs/install.md) and the [EKS docs](https://docs.aws.amazon.com/eks/latest/userguide/mesh-k8s-integration.html). We could add a generation script, or modify both to reference each yaml file in deploy individually, or remove the individual files and only update `all.yaml`.

To keep it simple, i think we can get rid of individual files in favor of all.yaml. Will follow up in a separate PR if that is ok.

nckturner:
Sounds good.

Issue #63, opened 2019-09-08
v0.1.2: unable to create Virtual nodes or Virtual routers


fmedery:
The latest release does not create Virtual Routers and Virtual Services.
Error in the containers logs:

```
E0908 19:05:20.307544       1 controller.go:443] error syncing 'appmesh-demo/colorteller-black': error updating virtual node status: Operation cannot be fulfilled on virtualnodes.appmesh.k8s.aws "colorteller-black": the object has been modified; please apply your changes to the latest version and try again

I0908 19:05:20.442223       1 virtualnode.go:96] Created virtual node colorteller-blue-appmesh-demo

E0908 19:05:20.462389       1 controller.go:443] error syncing 'appmesh-demo/colorteller-blue': error updating virtual node status: Operation cannot be fulfilled on virtualnodes.appmesh.k8s.aws "colorteller-blue": the object has been modified; please apply your changes to the latest version and try again
(...)
E0908 19:07:49.287511       1 controller.go:443] error syncing 'appmesh-demo/colorteller.appmesh-demo': error creating virtual router: InvalidParameter: 1 validation error(s) found.
```

nckturner:
The latest release was modified to run 5 goroutines instead of 1, so it could be possible you're seeing contention between them?  Are the objects not being created successfully?  I see `Created virtual node` in your logs.

stefanprodan:
@nckturner I had the same problem with Flagger, this is how I solved it https://github.com/weaveworks/flagger/blob/master/pkg/canary/status.go#L70

nckturner:
Thanks @stefanprodan, we should give that a try.

fmedery:
> The latest release was modified to run 5 goroutines instead of 1, so it could be possible you're seeing contention between them? Are the objects not being created successfully at all (ignoring the error messages)? I see `Created virtual node` in your logs.

@nckturner  No the objects were not created.

@stefanprodan tx for pointing out the workaround.

kiranmeduri:
Yes, I have seen the contention and corresponding error messages. @stefanprodan your fix should help in this case, but I believe controller code needs some TLC. 

@fmedery this contention happens when the status is being updated, however the resources should have been created.

kiranmeduri:
There is a non-backward compatible change in CRD where virtual-router now is required and need to provide a listener. This is the reason for 

```
[...]

PR #64, opened 2019-09-09
Using RetryOnConflict to deal with concurrent updates to resources

kiranmeduri:
*Issue #, if available:*
Related to #63 

*Description of changes:*
When updating status controller is not handling concurrent resource updates properly. There are two controller handlers that try to update a resource namely event handler and reconciler. This fix introduces retries to deal with this. Thanks to @stefanprodan for the reference to use RetryOnConflict (https://github.com/weaveworks/flagger/blob/master/pkg/canary/status.go)

### DIFF
--- a/pkg/controller/virtualnode.go
+++ b/pkg/controller/virtualnode.go
@@ -153,22 +153,23 @@ func (c *Controller) updateVNodeCondition(vnode *appmeshv1beta1.VirtualNode, con
 	now := metav1.Now()
 	if condition == (appmeshv1beta1.VirtualNodeCondition{}) {
 		// condition does not exist
-		condition = appmeshv1beta1.VirtualNodeCondition{
+		newCondition := appmeshv1beta1.VirtualNodeCondition{
 			Type:               conditionType,
 			Status:             status,
 			LastTransitionTime: &now,
 		}
+		vnode.Status.Conditions = append(vnode.Status.Conditions, newCondition)
 	} else {
 		// condition exists and not set to status
 		condition.Status = status
 		condition.LastTransitionTime = &now
 	}
 
-	err := c.setVirtualNodeStatusCondition(vnode, condition)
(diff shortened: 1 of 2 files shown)


pkg/controller/virtualnode.go
```diff
@@ -144,26 +145,45 @@ func (c *Controller) updateVNodeActive(vnode *appmeshv1beta1.VirtualNode, status
...
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		var getErr error
+		if !firstTry {
+			vnode, getErr = c.meshclientset.AppmeshV1beta1().VirtualNodes(vnode.Namespace).Get(vnode.GetName(), metav1.GetOptions{})
+			if getErr != nil {
+				return getErr
+			}
+		}
+		vnodeCopy := vnode.DeepCopy()
+		vnodeCopy.Status.Conditions = append(vnode.Status.Conditions, condition)
```
mogren: What happens here? On the first try we do a `Get` and overwrite the vnode parameter, but for the next retries we just do a deep-copy and keep appending the condition? 

Would it be possible to to the `Get` outside the retry loop? Why do we keep appending the same condition, can't we just do that once since only do one get? Same for `setVirtualNodeStatusCloudMapService()`

kiranmeduri: It is inverse, on first try we do update directly and if update fails we try to get latest version and override. Ideally we should detect that previous update is not related to same field, but it is not easy to do in all cases. Also the status is derived from truth in AWS service in this case, so we should be okay.

kiranmeduri: you are right about appending the condition, that is a bug. I will fix it. thanks

mogren: Ah, thanks for the explanation. I guess the name `firstTry` with the negation confused me. Something like `updateVnode = false`, then setting it to `true` after the first try would be easier to read.

PR #67, opened 2019-09-11
Disambiguate router names in example

bcelenza:
*Description of changes:*

The current example template referenced in https://github.com/aws/aws-app-mesh-examples/blob/master/walkthroughs/eks/base.md has an error that results in the colorteller virtual router mapping to the color gateway's route.

This change disambiguates the routers and routes by naming the routers accordingly.

Are router names intended to be optional in the controller? If so, there is a broader bug here that warrants a separate pull request.

*Testing:*
Followed the guide at the walkthrough link above, observed 500 error and invalid router to route mapping in App Mesh configuration. Updated with proposed changes and mapping works as expected.

### DIFF
--- a/examples/color.yaml.template
+++ b/examples/color.yaml.template
@@ -103,6 +103,7 @@ metadata:
 spec:
   meshName: ${MESH_NAME}
   virtualRouter:
+    name: color-router
     listeners:
       - portMapping:
           port: 9080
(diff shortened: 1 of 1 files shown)


nckturner:
@bcelenza yes they were intended to be optional.  I.e., if not specified default to service name.  Can you create an issue for the broader bug?

Issue #69, opened 2019-09-12
Virtual service name not used for virtual router


bcelenza:
If an explicit name is not provided as part of a virtual router configuration in the resource definition, the virtual service name should be used. However, this is not happening, so if multiple virtual services and routers are created, the virtual routers will share whichever route was created first by the controller.

See also #67 for a temporary fix in the existing example.

bcelenza:
Resolved with #72

PR #72, opened 2019-09-15
Use Virtual Service name for Virtual Router

bcelenza:
*Issue #, if available:* #69

*Description of changes:*

Fix for the behavior that should default to the Virtual Service's name if the Virtual Router is not given an explicit name.

*Testing:*

Tested via `make example`. I noticed the app mesh object does not have unit tests. Is the appropriate place [here](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/controller/virtualservice_test.go)?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,6 @@ LDFLAGS?="-X main.version=${VERSION} -X main.gitCommit=${GIT_COMMIT} -X main.bui
 GO111MODULE=on
 # Docker
 IMAGE=amazon/app-mesh-controller
-REGION=$(shell aws configure get region)
 REPO=$(AWS_ACCOUNT).dkr.ecr.$(AWS_REGION).amazonaws.com/$(IMAGE)
 VERSION=v0.1.2
 
--- a/examples/color.yaml.template
+++ b/examples/color.yaml.template
@@ -103,7 +103,6 @@ metadata:
 spec:
   meshName: ${MESH_NAME}
   virtualRouter:
-    name: color-router
     listeners:
       - portMapping:
           port: 9080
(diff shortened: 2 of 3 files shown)


nckturner:
>  I noticed the app mesh object does not have unit tests. 

Are you referring to the mesh struct here?

bcelenza:
> > I noticed the app mesh object does not have unit tests.
> 
> Are you referring to the mesh struct here?

Yes, all structs defined in appmesh.go

kiranmeduri:
Isn't this handled already @ https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/controller/virtualservice.go#L45. May be a bug in that logic there as https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/controller/virtualservice.go#L296 should handle it appropriately.

bcelenza:
> Isn't this handled already @ https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/controller/virtualservice.go#L45. May be a bug in that logic there as https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/controller/virtualservice.go#L296 should handle it appropriately.

It looks like the conditional check [here](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/controller/virtualservice.go#L298-L300) has the same issue.

As for where the correct place to put the fix is, I am open to suggestions here, but `virtualservice.go` might be a better place. Seems like we have duplicated this logic so I can remove it in the other area.

Makefile
```diff
@@ -5,7 +5,6 @@ LDFLAGS?="-X main.version=${VERSION} -X main.gitCommit=${GIT_COMMIT} -X main.bui
 GO111MODULE=on
 # Docker
 IMAGE=amazon/app-mesh-controller
-REGION=$(shell aws configure get region)
```
bcelenza: This is unused in the Makefile, and mislead me for a bit (my default region is different from my ENV var). Removing it.

PR #76, opened 2019-10-01
Add enum to httpRetryEvents

nckturner:
*Description of changes:*
* Add HTTP retry policy to the example.
* Add enum to `httpRetryEvents`.
* Fix docs formatting bug.
* Regenerate example (removes explicit router names).

### DIFF
--- a/deploy/all.yaml
+++ b/deploy/all.yaml
@@ -323,6 +323,11 @@ spec:
                             type: array
                             items:
                               type: string
+                              enum:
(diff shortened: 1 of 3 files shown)


kiranmeduri:
Added example here https://github.com/aws/aws-app-mesh-examples/pull/206

nckturner:
Ok, should we have it here too until we can get rid of the example completely?

kiranmeduri:
I think given that this is a special use-case I am inclined to direct users to aws-app-mesh-examples. We can leave the current basic example here until we completely deprecate.

PR #78, opened 2019-10-02
Release v0.2.0

nckturner:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,79 @@
+# CHANGELOG
+
+## v0.2.0
+
+### Summary
+
+This version introduces [http header matching](https://github.com/aws/aws-app-mesh-roadmap/issues/15) and [explicit priority to routes](https://github.com/aws/aws-app-mesh-roadmap/issues/77).
+
+### Changes
+
+* Show CircleCI status for master branch only (https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/92, @nckturner)
+* Add go fmt check to CI and fix code formatting (https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/89, @stefanprodan)
+* Fix virtual node condition message label (https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/88, @stefanprodan)
+* helm, eksctl install instructions (https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/84, @nckturner)
+* Status badge for circleCI (https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/87, @nckturner)
+* Add CircleCI config (https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/86, @stefanprodan)
+* Expose klog flags in cobra flag set (https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/85, @nckturner)
+* Add enum to httpRetryEvents (https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/76, @nckturner)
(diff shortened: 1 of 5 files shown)


docs/example.md
```diff
@@ -2,7 +2,7 @@
 
 After following the [install instructions](install.md), you can deploy an example application with:
 
-    curl https://raw.githubusercontent.com/aws/aws-app-mesh-controller-for-k8s/v0.1.2/examples/color.yaml | kubectl apply -f -
+    curl https://raw.githubusercontent.com/aws/aws-app-mesh-controller-for-k8s/master/examples/color.yaml | kubectl apply -f -
```
mogren: Just curious, why not just
```
kubectl apply -f https://raw.githubusercontent.com/aws/aws-app-mesh-controller-for-k8s/master/examples/color.yaml
```

nckturner: Probably cause I didn't know you could do that.

CHANGELOG.md
```diff
@@ -0,0 +1,62 @@
+# CHANGELOG
+
+## v0.2.0
+
+### Summary
+
+This version introduces [http header matching](https://github.com/aws/aws-app-mesh-roadmap/issues/15) and [explicit priority to routes](https://github.com/aws/aws-app-mesh-roadmap/issues/77), and is a minor version release to reflect the changes both in this release and the previous.
```
tabern: the previous what? Should we be specific? "previous release 0.1.x"

tabern: what about cloudmap?

tabern: It's not super clear what the benefits are from this description. Is there a bigger change we should be making somewhere else?

nckturner: Cloud Map support was introduced in the previous version and I couldn't think of an eloquent way to say "the previous version should have been a minor version bump".

nckturner: Cloud Map support was introduced in the previous version and I couldn't think of an eloquent way to say "the previous version should have been a minor version bump".

nckturner:
> Not super clear the benefit here. We should be a bit more precise in the changelog. Are there also readme changes? Seems like some of that is missing.

@tabern The summary is just the major features/changes/fixes, and the list of changes includes everything other than minor docs tweaks.  What do you think is missing?  I can add more to the summary, its definitely a bit light right now.

CHANGELOG.md
```diff
@@ -0,0 +1,62 @@
...
+### Summary
+
+This version introduces [http header matching](https://github.com/aws/aws-app-mesh-roadmap/issues/15) and [explicit priority to routes](https://github.com/aws/aws-app-mesh-roadmap/issues/77).
+
+### Changes
+
+* Update installation instruction for aws-app-mesh-inject (https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/60, @fmedery)
+* Fix unit tests and added Makefile target (https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/61, @kiranmeduri)
+* Using RetryOnConflict to deal with concurrent updates to resources (https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/64, @kiranmeduri)
+* Add http-header match and priority to routes (https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/56, @kiranmeduri)
```
kiranmeduri: did retry-policy not make the cut?

bcelenza: Same question for the virtual router name fix.

nckturner: I thought retry was there, but anyway I'll just add everything since my subjective picking and choosing what's important is a bad idea.

kiranmeduri: No worries, it is still a massive improvement than what we have. Thanks for compiling the list.

Issue #80, opened 2019-10-02
Feature: Add go profiling support


stefanprodan:
To debug memory or TCP connections leaks we should enable the go profiler under `/debug`.

Issue #79, opened 2019-10-02
Bug: The web server can't be reached inside Kubernetes


stefanprodan:
The server binds on localhost only and Kubernetes probes or Prometheus can't access it.

PR #81, opened 2019-10-02
Fix Prometheus scraping, enable pprof and update to go 1.13

stefanprodan:
This PR binds the controller server on all IPs so that Prometheus can scrape the `/metrics` endpoint and enables profiling:

```sh
kubectl -n appmesh-system port-forward deploy/appmesh-controller 10555:10555

go tool pprof http://localhost:10555/debug/pprof/profile
```

Fix: #79
Fix: #80
Fix: #83

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,7 @@
 module github.com/aws/aws-app-mesh-controller-for-k8s
 
+go 1.13
+
 require (
 	github.com/BurntSushi/toml v0.3.1 // indirect
 	github.com/aws/aws-sdk-go v1.23.18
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/aws/aws-app-mesh-controller-for-k8s
 
+go 1.13
```
jaypipes: Was this required for net/http/pprof support?

stefanprodan: No it's not required but pinning the go version in go.mod is good practice in my opinion. There were many problems with go 1.12 and modules, upgrading to 1.13 is a safe bet :)

jaypipes: Got it. Thanks! :)

stefanprodan:
One this is merged and released scraping will work as specified in https://github.com/aws/eks-charts/pull/1

PR #85, opened 2019-10-07
Expose klog flags in cobra flag set

nckturner:
*Description of changes:*
Currently there is no way of setting logging verbosity.  This PR fixes that by exposing klog flags.

### DIFF
--- a/cmd/app-mesh-controller/root.go
+++ b/cmd/app-mesh-controller/root.go
@@ -1,12 +1,14 @@
 package main
 
 import (
-	"flag"
+	goflag "flag"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


cmd/app-mesh-controller/root.go
```diff
@@ -150,3 +161,19 @@ func getConfig() (controllerConfig, error) {
 		},
 	}, nil
 }
+
+// addKlogFlags adds flags from k8s.io/klog
+func addKlogFlags(fs *flag.FlagSet) {
```
nckturner: This was borrowed from the kubernetes `globalflag` package: https://github.com/kubernetes/component-base/blob/kubernetes-1.15.0/cli/globalflag/globalflags.go#L41-L54

PR #88, opened 2019-10-14
Fix virtual node condition message label

stefanprodan:
This PR fixes the `VirtualNodeCondition.Message` JSON name that was mislabeled as `reason` thus overriding the  `VirtualNodeCondition.Reason` value.

### DIFF
--- a/pkg/apis/appmesh/v1beta1/types.go
+++ b/pkg/apis/appmesh/v1beta1/types.go
@@ -124,48 +124,48 @@ type Route struct {
 }
 
 type HttpRoute struct {
-	Match       HttpRouteMatch   `json:"match"`
-	Action      HttpRouteAction  `json:"action"`
+	Match  HttpRouteMatch  `json:"match"`
+	Action HttpRouteAction `json:"action"`
 	// +optional
 	RetryPolicy *HttpRetryPolicy `json:"retryPolicy,omitempty"`
 }
 
 type HttpRouteMatch struct {
-	Prefix  string            `json:"prefix"`
+	Prefix string `json:"prefix"`
 	// +optional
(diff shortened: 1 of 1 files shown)


pkg/apis/appmesh/v1beta1/types.go
```diff
@@ -399,7 +399,7 @@ type VirtualNodeCondition struct {
 	Reason *string `json:"reason,omitempty"`
 	// A human readable message indicating details about the transition.
 	// +optional
-	Message *string `json:"reason,omitempty"`
+	Message *string `json:"message,omitempty"`
```
jaypipes: @stefanprodan AFAICT, this is the only attribute that substantively was changed, yes? All the other changes are result of go fmt, yeah?

stefanprodan: Yes, my editor (Goland) runs go fmt before a commit. In the future we should add a fmt check job to CI and validate that all go files are properly formatted, I suspect many aren't.

stefanprodan:
@jaypipes I've renamed the commit and PR description.

pkg/apis/appmesh/v1beta1/types.go
mogren: In the AWS CNI we have a `check-format` target [in the Makefile](https://github.com/aws/amazon-vpc-cni-k8s/blob/8f40e8ebfb0a00f6a7330fe64f923b926dec9669/Makefile#L120-L134), and then just call that target in the [`.travis.yml`](https://github.com/aws/amazon-vpc-cni-k8s/blob/8f40e8ebfb0a00f6a7330fe64f923b926dec9669/.travis.yml#L26)

Issue #98, opened 2019-11-06
Prometheus instrumentation 


stefanprodan:
The controller should export metrics for App Mesh API calls, meshes, virtual nodes and virtual services.

Issue #96, opened 2019-11-06
TooManyRequestsException being triggered


lavignes:
I've been seeing `TooManyRequestsException`s being returned from App Mesh  frequently in the controller logs.

What sort of rate-limiting or back-off does the controller do? Is it configurable?

stefanprodan:
There is no rate-limiting in place as far as I'm aware, the controller reconciles all objects found in the cluster every [30s](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/cmd/app-mesh-controller/root.go#L110) and does an API call to App Mesh for each object from [5 threads](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/controller/controller.go#L31). I think the controller should detect changes in the Kubernetes custom resources and call the App Mesh API only for those objects. This can be achieved using the `cmp` package, [here](https://github.com/weaveworks/flagger/blob/master/pkg/controller/controller.go#L127) is an example. Right now if you have 100 virtual nodes in a cluster, the controller will do 100 POST requests to App Mesh even if none of the virtual nodes have changed since the last sync. 

PS. I'm planning to instrument the controller and expose Prometheus metrics for the API requests. We will be able to monitor the req/sec rate, error rate and latency between the controller and App Mesh API to better undestand where the problem is.

stefanprodan:
After instrumenting the App Mesh API calls this is what's happening 500 req/min with just a handful of objects (5 nodes and 2 virtual service) 😱 

<img width="1462" alt="Screenshot 2019-11-06 at 11 18 10" src="https://user-images.githubusercontent.com/3797675/68285141-66975480-0087-11ea-8240-6b66ce7ab463.png">

nckturner:
@lavignes how many App Mesh objects do you have in your cluster?

nckturner:
>  I think the controller should detect changes in the Kubernetes custom resources and call the App Mesh API only for those objects. 

I think we also want to overwrite objects that are modified in the App Mesh API as soon as possible so we need to temporarily query the API regardless of whether the custom resource changed. (Doesn't have to be every time, of course).

dastbe:
Rate limiting in the controller gets rid of the exceptions, but I don’t think that gets to what the customer experience should be here and what scale we should be designing for. If I get 10 tps to App Mesh, then I can just barely describe 300 resources every 30s (not even considering mutation for now).

So let’s think about, for a 30s period

[...]

PR #99, opened 2019-11-06
Implement Prometheus instrumentation 

stefanprodan:
This PR adds instrumentation for the controller App Mesh API calls and exports Prometheus metrics for meshes, virtual nodes and virtual services.

Metrics:
- `appmesh_mesh_state{name}` gauge
- `appmesh_virtual_node_state{name, mesh}` gauge
- `appmesh_virtual_service_state{name, mesh}`  gauge
- `appmesh_api_request_duration_seconds{kind, object, operation}`  histogram

The controller instrumentationr records mesh, virtual node and virtual service operations as gauges. For each object the gauge value represents the current state, `1` means that the object is active while `0` means that the object has been deleted.

The API client instrumentation records the duration of App Mesh API calls based on object kind, name and operation type. The operation type can be get, create, update or delete. The object kind can be mesh, virtual node, virtual route, virtual router or virtual service. 

The `appmesh_api_request_duration_seconds` histogram helps track issues like #96 

Fix: #98 

### DIFF
--- a/pkg/aws/appmesh.go
+++ b/pkg/aws/appmesh.go
@@ -76,6 +76,11 @@ func (v *Mesh) Name() string {
 
 // GetMesh calls describe mesh.
 func (c *Cloud) GetMesh(ctx context.Context, name string) (*Mesh, error) {
+	begin := time.Now()
+	defer func() {
+		c.stats.SetRequestDuration("mesh", name, "get", time.Since(begin))
+	}()
+
 	ctx, cancel := context.WithTimeout(ctx, time.Second*DescribeMeshTimeout)
 	defer cancel()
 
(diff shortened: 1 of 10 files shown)


pkg/aws/appmesh.go
```diff
@@ -231,6 +246,11 @@ func (v *VirtualNode) BackendsSet() set.Set {
 
 // GetVirtualNode calls describe virtual node.
 func (c *Cloud) GetVirtualNode(ctx context.Context, name string, meshName string) (*VirtualNode, error) {
+	begin := time.Now()
+	defer func() {
+		c.stats.SetRequestDuration("virtual_node", name, "get", time.Since(begin))
```
kiranmeduri: is it better to normalize? `<meshName>.<virtualNodeName>`

stefanprodan: I would not use dots in Prometheus metrics, we could add the mesh a separate label if you want.

nckturner: Agreed with @stefanprodan, let's add as a separate label.

stefanprodan: I would like to do this and add more instrumentation in a followup PR. The metrics as they are now are sufficient for the Grafana dashboard I made in https://github.com/aws/eks-charts/pull/30

nckturner:
Can you rebase?

stefanprodan:
@nckturner rebase done

PR #114, opened 2019-12-17
Allow using Cloud Map Http Namespace

kiranmeduri:
*Issue #, if available:*

*Description of changes:*
Previously we disallowed this due the concern that DNS will not resolve. However, service name resolution is decoupled from where virtual-nodes are registered w.r.t service-discovery. Removing this check and allowing HTTP namespace support. Note that the examples still recommend using PrivateDNSNamespace. See https://github.com/kiranmeduri/aws-app-mesh-examples/commit/029bed2ea83cdc5636e41cccdb7dbbe86d9f52df on how to use Kubernetes DNS with cloudmap HTTP namespace support.


nckturner:
Nice!  This makes sense, since one can substitute it with any other DNS to allow the app to perform the initial resolution, and if we're using cloudmap service discovery then nothing else is needed.

Issue #113, opened 2019-12-14
ColorApp Question: What's the purpose of the colorgateway VirtualService


eeeschwartz:
As a new App Mesh user, I spend a lot of time poring over the ColorApp example. I'm struggling to understand the purpose of the [colorgateway VirtualService](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/examples/color.yaml#L125).

It seems that VirtualService routes are only relevant for communication between VirtualNodes within a mesh. Correct so far?

In this case, why define a VirtualService for the colorgateway since it only receives traffic from outside of the mesh. At least that's how it's described in the [EKS walkthrough in the aws-app-mesh-examples repo](https://github.com/aws/aws-app-mesh-examples/blob/master/walkthroughs/eks/base.md#the-application). The curler pod makes requests to the colorgateway from outside of the mesh (i.e. it's deployed without an exposed port so no sidecar; also doesn't define a VirtualNode).

When you `kubectl exec curler-dc65c8c79-zbtkb -- curl colorgateway:9080/color`, the VirtualService config does not have an effect. For instance, I changed it to route traffic to `some-other-virtual-node` but it had no effect: 

```
apiVersion: appmesh.k8s.aws/v1beta1
kind: VirtualService
metadata:
  name: colorgateway.appmesh-demo
  namespace: appmesh-demo
spec:
  meshName: color-mesh
  virtualRouter:
    listeners:
      - portMapping:
          port: 9080
          protocol: http
  routes:
    - name: gateway-route
      http:
        match:
          prefix: /color
        action:
          weightedTargets:
            - virtualNodeName: some-other-virtual-node
              weight: 1
```

`kubectl exec curler-dc65c8c79-zbtkb -- curl colorgateway:9080/color` still routed to the `colorgateway` service as before.

Please let me know what I'm missing. Thanks in advance.

kiranmeduri:
You are right, it is not needed and can be dropped.

lavignes:
It looks like this issue can be closed. Feel free to re-open it if you need more clarification on this topic. :)

PR #115, opened 2019-12-17
Remove colorgateway VirtualService

eeeschwartz:
*Issue #, if available:*

#113 

*Description of changes:*

Remove colorgateway VirtualService so new users don't mistake App Mesh as supporting ingress routing (though it [may someday](https://github.com/aws/aws-app-mesh-roadmap/issues/37))

### DIFF
--- a/docs/example.md
+++ b/docs/example.md
@@ -50,7 +50,6 @@ You should see a collection of virtual services, virtual nodes and a mesh, along
     virtualnode.appmesh.k8s.aws/colorteller-blue-appmesh-demo    3h
     virtualnode.appmesh.k8s.aws/colorteller-red-appmesh-demo     3h
     NAME                                                                         AGE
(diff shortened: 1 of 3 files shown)


nckturner:
@kiranmeduri @eeeschwartz I guess the other option would be to add a virtual node to the curler pod that we launch with the demo, that way this is used.  I do think having the virtualservice makes for a more complete demo, because if this was real, you'd have other services talking to colorteller and you'd need it.  But I think this makes sense, as you pointed out, we should put an ingress here once it exists, and it might be confusing for users.

eeeschwartz:
Thanks @nckturner that makes sense. FWIW I realize that my confusion was also the 'colorgateway' naming. It imported a bunch of assumptions based on Istio Gateways and their role at the edge of the mesh.

PR #116, opened 2019-12-26
Fixing access to cloudmap IPs refresh

athilaelhuno:
*Issue #,Fixing access to Cloudmap IPs refresh:*
The IPs are not updated when the PODs are restarting or recreated
*Description of changes:*
Update the service discovery access in the main policy


nckturner:
> The IPs are not updated when the PODs are restarting or recreated

Why are the additional servicediscovery permissions required for pod IPs being kept up to date?

kiranmeduri:
> > The IPs are not updated when the PODs are restarting or recreated
> 
> Why are the additional servicediscovery permissions required for pod IPs being kept up to date?

Those are required to keep everything in sync. @athilaelhuno thanks for the fix.

PR #119, opened 2020-01-14
Stop using status field to sync cloudmap service for virtual-node

kiranmeduri:
*Issue #, if available:*
- https://github.com/aws/aws-app-mesh-roadmap/issues/146

*Description of changes:*
(TIL) Using status field to decide whether to take action or not is an anti-pattern. This is causing bug where, if user deletes a Service on Cloud Map side using AWS tooling, then controller is not creating the service again. This results in pods which do not get registered with Cloud Map due to stale service-id. AWS Cloud Map APIs do not fail synchronously i.e. RegisterInstance will not fail if service is missing, instead it returns an Operation response that will eventually fail.

This change ignores the service-id in status field (informational purposes only) and calls CreateService. Note that it does not call GetService because CreateService is cheaper.


nckturner:
>  if user deletes a Service on Cloud Map side using AWS tooling, then controller is not creating the service again. 

This is a good catch.  Makes sense we should not be using status.  If we want to reduce calls we can cache something and expire it. 

> Note that it does not call GetService because CreateService is cheaper.

Makes sense, I'm guessing you verified this with the cloudmap team?

kiranmeduri:
> Makes sense, I'm guessing you verified this with the cloudmap team?

Not yet, it is based on my observation. Moreover getService requires serviceID instead of serviceName. So one has to scan all the services and find the appropriate service.

PR #121, opened 2020-01-16
Added metrics to control loop functions

kiranmeduri:
*Issue #, if available:*
N/A
*Description of changes:*
Added more metrics to track control-loop and external api calls (cloudmap).


kiranmeduri:
> I would use the same subsystem `appmesh` and rename the metric to `operation_duration_seconds`. Using the `appmesh` suffix tells users that these metrics are generated by the AppMesh controller.

I wondered about that and initially used the same subsystem, but changed it to be standalone. I will revert it to use appmesh prefix. Any reason not to use _namespace_ instead of _subsystem_?

pkg/metrics/recorder.go
```diff
@@ -93,3 +111,9 @@ func (r *Recorder) SetVirtualServiceInactive(name string, mesh string) {
 func (r *Recorder) SetRequestDuration(kind string, object string, operation string, duration time.Duration) {
 	r.apiRequestDuration.WithLabelValues(kind, object, operation).Observe(duration.Seconds())
 }
+
+// RecordOperation records the duration of operation (e.g. API Call, Function exection)
+// based on object kind, name and operation type. The operation type can be get, create, update, delete
```
nckturner: > The operation type can be get, create, update, delete

In the instances in this PR you are using function names, do you want to make this comment reflect those usages?  Also object is always empty, when would it be used?

kiranmeduri: > Also object is always empty, when would it be used?

object is not always empty, for e.g. 
```
pkg/aws/cloudmap.go:            c.stats.RecordOperation("cloudmap", "service", "create", time.Since(begin))
pkg/aws/cloudmap.go:            c.stats.RecordOperation("cloudmap", "service", "getByID", time.Since(begin))
```

kiranmeduri:
Testing done

```
$ curl http://<controller>:10555/metrics
...
# HELP appmesh_controller_error Errors in controller
# TYPE appmesh_controller_error counter
appmesh_controller_error{controller="pod",errorcode="register_instance"} 84
...
# HELP appmesh_operation_duration_seconds Seconds spent performing operation.
# TYPE appmesh_operation_duration_seconds histogram
appmesh_operation_duration_seconds_bucket{kind="cloudmap",object="instance",operation="list",le="0.005"} 0
appmesh_operation_duration_seconds_bucket{kind="cloudmap",object="instance",operation="list",le="0.01"} 0
appmesh_operation_duration_seconds_bucket{kind="cloudmap",object="instance",operation="list",le="0.025"} 1664
...
appmesh_operation_duration_seconds_bucket{kind="podctl",object="",operation="handlePod",le="0.005"} 8359
appmesh_operation_duration_seconds_bucket{kind="podctl",object="",operation="handlePod",le="0.01"} 8465
...
```

Issue #107, opened 2019-11-25
Instances are not removed from Cloud Map when deleting the virtualNode


bemer:
If you create a VirtualNode in the kubernetes with `spec.serviceDiscovery.cloudMap` the controller register the Pods IP Addresses within the Cloud Map namespace, but when you delete the virtualNode the pods are still registered in the Cloud Map namespace.

I was using this VirtualNode spec:

```
apiVersion: appmesh.k8s.aws/v1beta1
kind: VirtualNode
metadata:
  name: nodejs-app
  namespace: appmesh-workshop-ns
spec:
  meshName: appmesh-workshop
  listeners:
    - portMapping:
        port: 3000
        protocol: http
  serviceDiscovery:
    cloudMap:
      namespaceName: appmeshworkshop.pvt.local
      serviceName: nodejs
```

The way to get the instances deregistered from Cloud Map was using the Cloud Map api: 

```
SERVICE_ID=$(aws servicediscovery list-services --filters Name="NAMESPACE_ID",Values=$NAMESPACE,Condition="EQ" | jq -r ' .Services[] | [ .Id ] | @tsv ' )
aws servicediscovery list-instances --service-id $SERVICE_ID | jq -r ' .Instances[] | [ .Id ] | @tsv ' |\
  while IFS=$'\t' read -r instanceId; do 
    aws servicediscovery deregister-instance --service-id $SERVICE_ID --instance-id $instanceId
  done
```

nckturner:
I can confirm this, it looks like we do not have logic for cleaning up the instances and the cloudmap service when the virtualnode is deleted.  Arguably not a bug, this was intended as a follow up task from the original Cloud Map support PR, but it looks like it fell through the cracks. As a side note, when pods are deleted, Individual instance deregistration does work.  

If we can know that the virtualnode is 1:1 with the Cloud Map service, then we should attempt to delete it.  Regardless, we should deregister instances from the service when the virtual node is deleted.  We do need a strategy for keeping track of these secondary resources though, that are not explicitly identified by a custom resource, but we still create.

lavignes:
> If we can know that the virtualnode is 1:1 with the Cloud Map service

A virtual node can select a subset of instances in a CloudMap service via Attributes. So many virtual nodes can point to instances of the same service.

nckturner:
@lavignes so in ECS, how do the instances get registered?  Service discovery isn't useful unless the endpoints are correct.

[...]

PR #122, opened 2020-01-16
Deregister instances from Cloud Map on virtualNode deletion

kiranmeduri:
*Issue #, if available:*
#107 

*Description of changes:*


pkg/controller/virtualnode.go
```diff
@@ -314,6 +314,9 @@ func vnodeLoggingNeedsUpdate(desired *appmeshv1beta1.VirtualNode, target *aws.Vi
 
 func (c *Controller) handleVNodeDelete(ctx context.Context, vnode *appmeshv1beta1.VirtualNode, copy *appmeshv1beta1.VirtualNode) error {
 	if yes, _ := containsFinalizer(vnode, virtualNodeDeletionFinalizerName); yes {
+		if err := c.deregisterInstancesForVirtualNode(ctx, vnode); err != nil {
+			return err
```
nckturner: Should we still try to delete the virtual node even if instance deregistration fails?  I.e. if cloudmap is down, is it safer to fail and retry on deregistering instances, and leave the virtualnode around until we succeed, or delete the virtualnode immediately without worrying about cleaning up cloudmap?  I'm thinking the former is correct, which is the logic you have, otherwise we potentially leave cloudmap instances around after the virtualnode is gone.

kiranmeduri: yeah former is safer and predictable.

PR #123, opened 2020-01-23
fix example to work with iam-for-pods

M00nF1sh:
Fix the example to work with iam-for-pods.

The envoy controller **is not running as root**, thus requires the fsGroup: 1337 setting to be able to read credentials mounted by iam-for-pods, otherwise, it just fails silently without obvious error logs.

Add this `fsGroup: 1337` setting will make this example works for both EC2 node(with/without iam-for-pods) and Fargate(requires iam-for-pods)

reference: https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-technical-overview.html#pod-configuration.

### DIFF
--- a/examples/color.yaml
+++ b/examples/color.yaml
@@ -152,6 +152,8 @@ spec:
         app: colorgateway
         version: v1
     spec:
+      securityContext:
+        fsGroup: 1337
       containers:
         - name: colorgateway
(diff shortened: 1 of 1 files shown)


nckturner:
@micahhausler double checking this is an expected requirement for application using IAM for pods.

PR #129, opened 2020-02-15
Initial version of development guide

kiranmeduri:
*Issue #, if available:*

*Description of changes:*


Makefile
```diff
@@ -7,6 +7,8 @@ GO111MODULE=on
 IMAGE=amazon/app-mesh-controller
 REPO=$(AWS_ACCOUNT).dkr.ecr.$(AWS_REGION).amazonaws.com/$(IMAGE)
 VERSION=v0.3.0
+GIT_COMMIT_SHORT=$(shell git log -1 --format=%h)
+DEV_VERSION=${VERSION}-${GIT_COMMIT_SHORT}
```
nckturner: I think its better to use to include information about recent tag and dirty working tree:
```
$ git describe --dirty --tags
v0.0.10-22-gcee18f5-dirty
```

I guess the only potential issue is using VERSION vs the most recent git tag, but I think it should be ok referring to the most recent git tag which should be present due to the github release.

Makefile
```diff
@@ -34,19 +36,16 @@ verify-codegen:
...
 	docker build -t $(IMAGE):$(VERSION) .
 
 .PHONY: push
 push:
-ifeq ($(AWS_ACCOUNT),)
-	$(error AWS_ACCOUNT is not set)
-endif
-	docker tag $(IMAGE):latest $(REPO):latest
-	docker push $(REPO):latest
+	docker tag $(IMAGE):$(DEV_VERSION) $(REPO):$(DEV_VERSION)
```
nckturner: Why remove the AWS_ACCOUNT check?  Its required in by REPO.

docs/development.md
```diff
@@ -0,0 +1,34 @@
...
+```
+make image push
+```
+- Deploy latest changes to appmesh-controller
+```
+make deploy
+```
+- Use examples from [aws-app-mesh-examples](https://github.com/aws/aws-app-mesh-examples/tree/master/walkthroughs) to verify the controller behavior.
+
+## Updating App Mesh CRD
```
efe-selcuk: Should `make go-fmt` also be added here?

Makefile
kiranmeduri: added it back. should we add same check to push-release as it also depends on REPO?

Makefile
kiranmeduri: TIL, thanks for that cmd.

docs/development.md
kiranmeduri: added. We should probably make it as pre-commit hooks.

PR #130, opened 2020-02-18
Add support for ACM/File listener TLS, backend TLS client policies, backend defaults

efe-selcuk:
*Issue #, if available:*

- https://github.com/aws/aws-app-mesh-roadmap/issues/38
- https://github.com/aws/aws-app-mesh-roadmap/issues/39

*Description of changes:*

Note: Currently affected by https://github.com/aws/aws-app-mesh-controller-for-k8s/issues/133

This adds support for ACM- and File-based listener TLS, backend TLS client policies, and backend defaults for TLS client policies. This is currently a basic pass-through of CRD fields to the SDK.

- [x] Update `aws-go-sdk` in go.mod to use the latest types from App Mesh
- [x] Update CRD schema in `deploy/all.yaml`
- [x] Update CRD structs in `pkg/apis/appmesh/v1beta1/types.go`
- [x] Update deepcopy functions using `make code-gen`
- [x] Update App Mesh client wrapper `pkg/aws/appmesh.go`
- [x] Update controller(s) under `pkg/controller/`


pkg/aws/appmesh.go
```diff
@@ -1487,3 +1517,34 @@ func durationToMillis(d *appmesh.Duration) *int64 {
...
 	return nil
 }
+
+func convertClientPolicy(c *appmeshv1beta1.ClientPolicy) *appmesh.ClientPolicy {
+	if c == nil {
+		return nil
+	}
+	clientPolicy := &appmesh.ClientPolicy{}
+	if c.Tls != nil {
+		tls := appmesh.ClientPolicyTls{
```
bcelenza: Could we be consistent on usage of pointers vs. values? Here we store `clientPolicy` as by pointer and  `tls` within it by value and reference the pointer later. I think storing the pointer is preferred.

pkg/apis/appmesh/v1beta1/types.go
```diff
@@ -453,6 +464,69 @@ type CloudMapServiceStatus struct {
...
 
+// General TLS Types
+
+type TlsValidationContext struct {
+	Trust TlsValidationContextTrust `json:"trust"`
+}
+
+type TlsValidationContextTrust struct {
+	Acm  *TlsValidationContextAcmTrust  `json:"acm,omitempty"`
+	File *TlsValidationContextFileTrust `json:"acm,omitempty"`
```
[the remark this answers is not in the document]
efe-selcuk: I figured if someone defined e.g. `File:` but didn't put any content, it shouldn't even be deserialized. Should we pass it through and let the API do the validation? I could follow that, so it's not misleading on what's in the CRD yaml vs what actually gets sent?

pkg/aws/appmesh.go
efe-selcuk: I was struggling with this a bit. I think by-value is actually preferred from a safety (and performance) perspective? Most of these things are just getting passed around immutably once created.

Though given the nullability in the SDK models, that ends up forcing use of pointers for stuff like the `SetTls` call, or the returned pointer to the `clientPolicy` to the caller.

Could use another opinion.

pkg/aws/appmesh.go
```diff
@@ -349,6 +349,24 @@ func (c *Cloud) CreateVirtualNode(ctx context.Context, vnode *appmeshv1beta1.Vir
...
+					})
+				}
+				if listener.Tls.Certificate.File != nil {
+					appmeshCert.SetFile(&appmesh.ListenerTlsFileCertificate{
+						CertificateChain: listener.Tls.Certificate.File.CertificateChain,
+						PrivateKey:       listener.Tls.Certificate.File.PrivateKey,
+					})
+				}
+				appmeshListener.SetTls(&appmesh.ListenerTls{
+					Mode:        &listener.Tls.Mode,
```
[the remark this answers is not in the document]
efe-selcuk: The field is nullable, so it has to be a pointer. The SDK model is generated that way. I should be consistent and use the `aws.String()` helper that is used elsewhere, though.

pkg/apis/appmesh/v1beta1/types.go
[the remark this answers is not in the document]
bcelenza: Yeah I think we want the contract to be as clear as possible in the CRD.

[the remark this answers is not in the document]
efe-selcuk: To be fair, I don't think this would improve clarity? Unless you mean clarity as in "if you have a stray definition, the API will error at you and point it out".

Anyway, I'll switch it elsewhere too, though I think that would mean we'd want to remove `omitempty` entirely, yes? I won't take that here but hopefully a reviewer closer to this can chime in.

I also just realized I didn't use the `required` filed in the CRD yaml spec, so I gotta add that anyway.

pkg/apis/appmesh/v1beta1/types.go
```diff
@@ -453,6 +464,69 @@ type CloudMapServiceStatus struct {
...
+	CertificateChain *string `json:"certificateChain"`
+	PrivateKey       *string `json:"privateKey"`
+}
+
+// END Listener TLS Types
+
+// Client Policy Types
+
+type ClientPolicy struct {
+	Tls *ClientPolicyTls `json:"tls,omitempty"`
```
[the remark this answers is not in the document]
efe-selcuk: Done

pkg/apis/appmesh/v1beta1/types.go
```diff
@@ -453,6 +464,69 @@ type CloudMapServiceStatus struct {
...
+}
+
+type ClientPolicyTls struct {
+	// +optional
+	Enforce bool `json:"bool,omitempty"`
+	// +optional
+	Ports []*int64 `json:"ports,omitempty"`
+	// +optional
+	Validation *TlsValidationContext `json:"validation,omitempty"`
```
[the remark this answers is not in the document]
efe-selcuk: Updated

pkg/apis/appmesh/v1beta1/types.go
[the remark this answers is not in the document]
nckturner: Generally `// +optional` and `omitempty` are used together when the field is allowed to be left out.  If the field is required by the App Mesh API, then it shouldn't be omitempty.

pkg/aws/appmesh.go
nckturner: I think small, short-lived structs are preferred to be stored by value, unless a pointer is needed.  Since in this case, the pointer is needed for the SetTls call, I think whether or not you use a pointer up front is a judgement call based on what you think is more readable.

pkg/apis/appmesh/v1beta1/types.go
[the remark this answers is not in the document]
bcelenza: Okay — I don’t think this is a big deal to keep as-is.

pkg/aws/appmesh.go
efe-selcuk: Got it. I'll do a pass over and try to keep consistent to using values until a pointer is needed.

pkg/apis/appmesh/v1beta1/types.go
```diff
@@ -453,6 +464,69 @@ type CloudMapServiceStatus struct {
...
+type ListenerTls struct {
+	Mode        string                 `json:"mode"`
+	Certificate ListenerTlsCertificate `json:"certificate"`
+}
+
+type ListenerTlsCertificate struct {
+	// +optional
+	Acm *ListenerTlsAcmCertificate `json:"acm,omitempty"`
+	// +optional
+	File *ListenerTlsFileCertificate `json:"file,omitempty"`
```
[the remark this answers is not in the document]
efe-selcuk: Keeping this as-is per earlier convo.

pkg/aws/appmesh.go
bcelenza: Makes sense, let’s keep it as a value by now. This can easily be changed later.

pkg/aws/appmesh.go
```diff
@@ -224,25 +224,43 @@ func (v *VirtualNode) Listeners() []appmeshv1beta1.Listener {
 	}
 
 	var listeners = []appmeshv1beta1.Listener{}
-	for _, appmeshListener := range v.Data.Spec.Listeners {
+	for _, sdkListener := range v.Data.Spec.Listeners {
```
nckturner: I like the clarification of "sdk" here vs appmesh.

efe-selcuk: I'm glad! It took some restraint to not apply it in other code, but I'll look to sticking to that for what I touch.

efe-selcuk:
Current state is that the implementation of the feature is final. However as mentioned in the overview this is affected by #133. It can be reviewed as-is, though there may be changes depending on how we handle that issue.

I was able to test this with a spec like:

```
---
apiVersion: v1
kind: Namespace
metadata:
  labels:
  name: meshy
---
apiVersion: appmesh.k8s.aws/v1beta1
kind: Mesh
metadata:
  namespace: meshy
  name: tlsmesh
---
apiVersion: appmesh.k8s.aws/v1beta1
kind: VirtualNode
metadata:
  name: filetls
  namespace: meshy
spec:
  meshName: tlsmesh
  listeners:
    - portMapping:
        port: 8080
        protocol: http
      tls:
        mode: PERMISSIVE
        certificate:
          file:
            certificateChain: "/path/to/file.pem"
            privateKey: "/path/to/key.pem"
  backends:
    - virtualService:
        virtualServiceName: some.cool.svc
        clientPolicy:
          tls:
            # enforce: true
            ports: [8080, 9090]
            validation:
              trust:
                file:
                  certificateChain: "/path/to/chain.pem"
  backendDefaults:
    clientPolicy:
      tls:
        enforce: false
        ports: []
        validation:
          trust:
            file:
              certificateChain: "/path/to/chain.pem"
  serviceDiscovery:
    dns:
      hostName: front.local
---
apiVersion: appmesh.k8s.aws/v1beta1
kind: VirtualNode
metadata:
  name: acmtls
  namespace: meshy
spec:
  meshName: tlsmesh
  listeners:
    - portMapping:
        port: 8080
        protocol: http
      tls:
        mode: STRICT
        certificate:
          acm:
            certificateArn: "arn:aws:acm:us-east-2:326882241343:certificate/1c2fd791-fa5b-49ef-948d-7cd5c8f57228"
  backends:
    - virtualService:
        virtualServiceName: some.cool.svc
        clientPolicy:
          tls:
            validation:
              trust:
                acm:
                  certificateAuthorityArns:
                    - "arn:aws:acm-pca:us-east-2:326882241343:certificate-authority/351293b2-e08c-4985-a61b-fee6b287a942"
  backendDefaults:
    clientPolicy:
      tls:
        enforce: true
        ports: [1234, 5678]
        validation:
          trust:
            acm:
              certificateAuthorityArns:
                - "arn:aws:acm-pca:us-east-2:326882241343:certificate-authority/351293b2-e08c-4985-a61b-fee6b287a942"
  serviceDiscovery:
    dns:
      hostName: cool.local
```

`kubectl get` returns a match. And describing the nodes by the app mesh API returns matching results, except for some inconsistencies where the "default" value is returned by the API, e.g. true for `enforce`.

efe-selcuk:
We will likely merge this despite #133 though if PR is approved please hold off on merging.

bcelenza:
Tested backend policies, backend defaults, and TLS with ACM on one of my test clusters. Verified everything was configured correctly down to the Envoys. LGTM.

pkg/apis/appmesh/v1beta1/types.go
```diff
@@ -352,13 +352,17 @@ type VirtualNodeSpec struct {
...
+	// +optional
 	Logging *Logging `json:"logging,omitempty"`
 }
 
 type Listener struct {
 	PortMapping PortMapping `json:"portMapping"`
 	// +optional
 	HealthCheck *HealthCheckPolicy `json:"healthCheck,omitempty"`
+	// +optional
+	Tls *ListenerTls `json:"tls,omitempty"`
```
M00nF1sh: nit: It's better to use `TLS` as field name to follow k8's convention Ref: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md
```All letters in the acronym should have the same case, using the appropriate case for the situation```
TLS in Ingress spec is using `TLS`

Comment (efe-selcuk) [reply pkg/apis/appmesh/v1beta1/types.go]: Ok. Are you suggesting the field name is `TLS`? Orr the yaml/json? Or both?

> TLS in Ingress spec is using TLS

What field is this referring to?

Comment (bcelenza) [reply pkg/apis/appmesh/v1beta1/types.go]: Clarification: Ingress appears to actually have this as all lowercase: `tls`

https://kubernetes.io/docs/concepts/services-networking/ingress/#tls

Comment (M00nF1sh) [reply pkg/apis/appmesh/v1beta1/types.go]: @bcelenza FYI: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/networking/types.go#L264 🤣

Comment (bcelenza) [reply pkg/apis/appmesh/v1beta1/types.go]: Haha, well consistently aligned with the formal definition I guess.

I think it makes sense to keep it lower case, because that's what we've done for dns: https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/deploy/all.yaml#L156

Comment (M00nF1sh) [reply pkg/apis/appmesh/v1beta1/types.go]: @bcelenza the attribute is lower case. it's the field name in golang structure.

Comment (bcelenza) [reply pkg/apis/appmesh/v1beta1/types.go]: Ah, thanks for the clarification.

File: pkg/apis/appmesh/v1beta1/types.go [hunk 9]
```diff
@@ -453,6 +464,71 @@ type CloudMapServiceStatus struct {
...
+
+type TlsValidationContextTrust struct {
+	// +optional
+	Acm *TlsValidationContextAcmTrust `json:"acm,omitempty"`
+	// +optional
+	File *TlsValidationContextFileTrust `json:"file,omitempty"`
+}
+
+type TlsValidationContextAcmTrust struct {
+	CertificateAuthorityArns []*string `json:"certificateAuthorityArns"`
```
M00nF1sh: is there a reason for []*string? `[]string` is better to me, normally we only use *string to represent optional semantic. 

BTW, I know aws-sdk-go is a hell filled with pointers(and they are fixing it in aws-sdk-go-v2)

pkg/apis/appmesh/v1beta1/types.go
```diff
@@ -453,6 +464,71 @@ type CloudMapServiceStatus struct {
...
+type ClientPolicy struct {
+	// +optional
+	Tls *ClientPolicyTls `json:"tls,omitempty"`
+}
+
+type ClientPolicyTls struct {
+	// +optional
+	Enforce *bool `json:"enforce,omitempty"`
+	// +optional
+	Ports      []*int64             `json:"ports,omitempty"`
```
M00nF1sh: same here, `[]int64` will be better

pkg/aws/appmesh.go
```diff
@@ -220,65 +220,80 @@ func (v *VirtualNode) Listeners() []appmeshv1beta1.Listener {
...
-// BackendsSet returns a set of Backends defined for virtual-node
-func (v *VirtualNode) BackendsSet() set.Set {
-	backends := v.Backends()
-	s := set.NewSet()
-	for i := range backends {
-		s.Add(backends[i])
+// BackendDefaults converts from the SDK types into CRD types
+func (v *VirtualNode) BackendDefaults() *appmeshv1beta1.BackendDefaults {
+	if v.Data.Spec.BackendDefaults == nil {
+		return &appmeshv1beta1.BackendDefaults{}
```
M00nF1sh: nit, what's the semantic empty value for BackendDefaults? is it always `&appmeshv1beta1.BackendDefaults{}` or `nil`? currently it's fine since we will always do null-checks for BackendDefaults.ClientPolicy(since it's a pointer). But these unclearly defined empty value is typical a source a bugs.

pkg/apis/appmesh/v1beta1/types.go
efe-selcuk: Ok, so just changing the field name in the struct to `TLS`, got it.

pkg/apis/appmesh/v1beta1/types.go
efe-selcuk: I don't have a good reason, I think I was just going of some existing fields. There was an earlier discussion where I settled on exactly the point you're making, just didn't think to apply it to the slices.

pkg/aws/appmesh.go
efe-selcuk: Good catch. This should be nil. It's an optional field in the API.

pkg/controller/virtualnode.go
```diff
@@ -264,12 +264,7 @@ func vnodeNeedsUpdate(desired *appmeshv1beta1.VirtualNode, target *aws.VirtualNo
 
 	// This needs to be updated since AppMesh VN name isn't the same as k8s VN name.
 	if desired.Spec.Backends != nil {
-		desiredSet := set.NewSet()
-		for i := range desired.Spec.Backends {
-			desiredSet.Add(desired.Spec.Backends[i])
-		}
-		currSet := target.BackendsSet()
-		if !desiredSet.Equal(currSet) {
+		if !reflect.DeepEqual(desired.Spec.Backends, target.Backends()) {
```
M00nF1sh: will the AppMesh api *guarantee* the backends is ordered using same order when it's created with? guess that's the original intention to use set compare

pkg/controller/virtualnode.go
```diff
@@ -279,6 +274,17 @@ func vnodeNeedsUpdate(desired *appmeshv1beta1.VirtualNode, target *aws.VirtualNo
 		}
 	}
 
+	if desired.Spec.BackendDefaults != nil {
+		if !reflect.DeepEqual(desired.Spec.BackendDefaults, target.BackendDefaults()) {
```
M00nF1sh: humm, my original comment on `func (v *VirtualNode) BackendDefaults() *appmeshv1beta1.BackendDefaults {` actually applies here now. ``` unclearly defined empty value is typically source of bugs. ```

`reflect.DeepEqual(nil, &object{}) == False`.

pkg/controller/virtualnode.go
```diff
@@ -279,6 +274,17 @@ func vnodeNeedsUpdate(desired *appmeshv1beta1.VirtualNode, target *aws.VirtualNo
 		}
 	}
 
+	if desired.Spec.BackendDefaults != nil {
```
M00nF1sh: if we fix `target.BackendDefaults() to return 'nil'` when `target.Data.Spec.BackendDefaults == nil`, this can be simplified to just be `!reflect.DeepEqual(desired.Spec.BackendDefaults, target.BackendDefaults())`

pkg/controller/virtualnode.go
```diff
@@ -512,6 +518,41 @@ func (c *Controller) mutateVirtualNodeForProcessing(vnode *appmeshv1beta1.Virtua
...
+			clientPolicy := vnode.Spec.BackendDefaults.ClientPolicy
+			if clientPolicy.Tls != nil {
+				mergeTlsClientPolicyDefaults(clientPolicy.Tls)
+			}
+		}
+	}
+}
+
+func mergeTlsClientPolicyDefaults(baseClientPolicyTls *appmeshv1beta1.ClientPolicyTls) {
+	if baseClientPolicyTls == nil {
```
M00nF1sh: redundant null check

efe-selcuk: I assume you're suggesting removing the null check at the call site, not here.

pkg/controller/virtualnode.go
efe-selcuk: I suspect it is stable since it's a paginated API, but I don't think there's a guarantee. That makes sense. I don't recall why I removed this (maybe I thought it was redundant), but I'll add it back.

pkg/controller/virtualnode.go
efe-selcuk: i.e. the call site becomes
```
if vnode.Spec.BackendDefaults != nil {
	mergeTlsClientPolicyDefaults(vnode.Spec.BackendDefaults.ClientPolicy)
}
```

Or would it be more appropriate to pass in a value and return a new copy?

M00nF1sh: i suggest remove it from function side. it's more like io.read(buff) where buff is expected to be a valid buffer.

pkg/controller/virtualnode.go
efe-selcuk: I think what confused me is that `Listeners` is not handled the same way, though that might be because we have a limit of 1 from the api right now.

pkg/aws/appmesh.go
efe-selcuk: I see why I did it, the other helpers like `Listeners()` and `Backends()` define it as an empty object/slice. Possibly due to API semantics.

pkg/apis/appmesh/v1beta1/types.go
efe-selcuk: > BTW, I know aws-sdk-go is a hell filled with pointers

And it seems they have conversion helpers for everything because of it.

efe-selcuk:
Forgot to put the BackendsSet logic back in, and now a couple TLS tests are failing. Working on it.

PR #137, opened 2020-03-12
Added E2E tests framework and test cases

M00nF1sh:
Added E2E tests framework and test cases.

## The tests cases is structured as follows:
* `/test/e2e/fishapp`: pods are composited by two container: colorApp and httpProxy. The colorApp is a container echo environment variable. the httpProxy allows we check pod's connectivity into other components within cluster. fishapp containers can be used as both client and server, so it can be structured to form either symmetrical mesh or asymmetrical mesh. (`fishapp` = `app by m00nf1sh` 😄 )
* `/test/e2e/fishapp/dynamic_stack`: a dynamic configurable symmetrical mesh based on fishapp. 
    * `$VirtualNodesCount` of VirtualNode
    *  `$VirtualServicesCount` of VirtualServices.
    *  each VirtualService have `$RoutesCountPerVirtualRouter`  of paths: `path-0`, `path-1`...
    *  each path above have `$TargetsCountPerRoute` targets with equal weight.
    *  each VirtualNode is allowed to talk to `$BackendsCountPerVirtualNode` of services.
    *  each VirtualNode have `$ReplicasPerVirtualNode` of pods.
    *  each pod have two container port, one is colorApp and intercepted by envoy used VirtualService backend. **Another is an HttpProxy un-intercepted by envoy(we use it to test pod's connectivity to services by doing portforwards xD**)
    * After the stack deploy succeeds, for each pod, we find it's corresponding node's allowed backend service, and request each path in these services for `$ConnectivityCheckPerURL` times.
* `test/e2e/fishapp/<other_stacks>`: **extension point** for other stacks based on fishapp.
* `test/e2e/<other_app>/<other_scenario>` :**extension point** for other application and scenarios.
* `test/e2e/framework`: contains support and utility code.

## current test workflow based on dynamic_stack(for both cloudMap & DNS):
1. Reset controller/injector into default ones **from latest helm**. (yes, it's a e2e test against our helm charts as well xD)
2. Install a "dynamic stack" called stackOld.
3. Verify stackOld's behaviors. (for each pod in this symmetrical mesh, check it's connectivity to other services allowed to access).
4. Upgrade controller/injector into new built ones from code **using helm**. (to check upgrade via helm works fine)
5. Verify stackOld's behaviors again. (to check new controller don't break existing application)
6. Install a new "dynamic stack" called stackNew.
7. Verify stackNew's behaviors.

## How to run tests manually
1. Run test against your cluster with controller/injector installed: `ginkgo -r -v test/e2e/ -- --kubeconfig=<kubeconfig> --cluster-name=<cluster-name> --aws-region=us-west-2 --aws-vpc-id=<vpc-id>`
2. Run test against your cluster with controller/injector installed and upgrade test against new controller image: `ginkgo -r -v test/e2e/ -- --kubeconfig=<kubeconfig> --cluster-name=<cluster-name> --aws-region=us-west-2 --aws-vpc-id=<vpc-id> --controller-image=<new-controller-image>`
3. Run test against your cluster with controller/injector installed and upgrade test against new injector image: `ginkgo -r -v test/e2e/ -- --kubeconfig=<kubeconfig> --cluster-name=<cluster-name> --aws-region=us-west-2 --aws-vpc-id=<vpc-id> --injector-image=<new-injector-image>`

## CircleCI run:
* https://app.circleci.com/pipelines/github/M00nF1sh/aws-app-mesh-controller-for-k8s/18/workflows/7552bc60-74a4-478b-84b0-ae0a1ca0c772/jobs/54

## Here is the plan to running test in CI for controller:

compile the controller and upload to ECR as controller: xxxxxx
create cluster with controller installed with image(controller: xxxxxx), and latest injector from helm chart. (with the help of this PR: aws/aws-k8s-tester#81), which allows to install controller & injector from latest helm chart, while allows to override controller/injector image.
launch the test suit against the cluster

## Here is the plan to running test in CI for injector:
compile the injector and upload to ECR as injector: xxxxxx
create cluster with injector installed with image(injector: xxxxxx), and latest controller from helm chart. (with the help of this PR: aws/aws-k8s-tester#81), which allows to install controller & injector from latest helm chart, while allows to override controller/injector image.
launch the test suit against the cluster


test/e2e/fishapp/shared/manifest_builder.go
```diff
@@ -0,0 +1,275 @@
...
+
+const (
+	appContainerPort       = 9080
+	httpProxyContainerPort = 8899
+)
+
+type ServiceDiscoveryType int
+
+const (
+	DNSServiceDiscovery ServiceDiscoveryType = iota
```
kiranmeduri: any reason not to use types in crd?

test/e2e/fishapp/shared/manifest_builder.go
```diff
@@ -0,0 +1,275 @@
...
+	appmeshv1beta1 "github.com/aws/aws-app-mesh-controller-for-k8s/pkg/apis/appmesh/v1beta1"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+const (
+	defaultAppImage       = "970805265562.dkr.ecr.us-west-2.amazonaws.com/colorteller:latest"
+	defaultHTTPProxyImage = "abhinavsingh/proxy.py:latest"
```
kiranmeduri: its not clear what the purpose of this is esp. the name may conflict with understanding of Envoy sidecar.

test/e2e/fishapp/shared/manifest_builder.go
```diff
@@ -0,0 +1,275 @@
...
+type ManifestBuilder struct {
+	MeshName             string
+	Namespace            string
+	ServiceDiscoveryType ServiceDiscoveryType
+
+	// required when serviceDiscoveryType == CloudMapServiceDiscovery
+	CloudMapNamespace string
+}
+
+func (b *ManifestBuilder) BuildNodeDeployment(instanceName string, replicas int32) *appsv1.Deployment {
```
kiranmeduri: will this rely on appmesh-injector running or should we remove that dependency?

test/e2e/framework/k8s/port_forward.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/onsi/ginkgo"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/portforward"
+	"k8s.io/client-go/transport/spdy"
+	"net/http"
+)
+
+// NewPortForwarder return a new port forwarder
```
kiranmeduri: Can you expand on this comment explaining how this is used by operator.

test/e2e/fishapp/shared/dynamic_stack.go
```diff
@@ -0,0 +1,631 @@
...
+	"sync"
+	"time"
+)
+
+const (
+	connectivityCheckRate = time.Second / 100
+	appMeshOperationRate  = time.Second * 1
+)
+
+// An dynamic generated stack designed to test app mesh integration :D
```
kiranmeduri: nit: An -> A

test/e2e/fishapp/shared/dynamic_stack.go
```diff
@@ -0,0 +1,631 @@
...
+	ReplicasPerVirtualNode int32
+
+	// how many time to check connectivity per URL
+	ConnectivityCheckPerURL int
+
+	// ====== runtime variables ======
+	mesh      *appmeshv1beta1.Mesh
+	namespace *corev1.Namespace
+
+	createdNodeVNs  []*appmeshv1beta1.VirtualNode
```
kiranmeduri: will it make it easier to track these as maps instead of slices?

test/e2e/fishapp/shared/manifest_builder.go
M00nF1sh: i only find a meshServiceDiscoveryType defined, but there is only DNS(no cloudMap).

test/e2e/fishapp/shared/manifest_builder.go
M00nF1sh: it's an http proxy server. this allows me to test whether a pod can reach a service programmatically.

e.g. code<running on mac/circleCI> access services through this HTTPProxy.

test/e2e/fishapp/shared/manifest_builder.go
M00nF1sh: yes, we rely on appmesh-injector. I don't think we should remove it, since the test is meant to test them as whole. BTW, we'll need controller when we test injector. 


Here is the plan to running test in CI for controller:
1. compile the controller and upload to ECR as controller: xxxxxx
2. create cluster with controller installed with image(controller: xxxxxx), and `latest injector` from helm chart. (with the help of this PR: https://github.com/aws/aws-k8s-tester/pull/81), which allows to install controller & injector from latest helm chart, while allows to override controller/injector image.
3. launch the test suit against the cluster
Here is the plan to running test in CI for injector:
1. compile the injector and upload to ECR as injector: xxxxxx
2. create cluster with injector installed with image(injector: xxxxxx), and `latest controller` from helm chart. (with the help of this PR: https://github.com/aws/aws-k8s-tester/pull/81), which allows to install controller & injector from latest helm chart, while allows to override controller/injector image.
3. launch the test suit against the cluster

test/e2e/framework/k8s/port_forward.go
M00nF1sh: e.g. To test whether pod(node-0-xxxx) can access service-1 from our code(which runs on mac or circleCI):

first, we setup a port-forward like localhost:8899:node-0-xxxx:8899. (httpProxy runs on 8899)
then we do curl -x "http://localhost:8899" "http://service-1:<service-port>/path-0"

test/e2e/fishapp/shared/dynamic_stack.go
M00nF1sh: thanks :D

test/e2e/fishapp/shared/dynamic_stack.go
M00nF1sh: maps works too, but slice is enough since all i need is `index -> node` :D

test/e2e/fishapp/dynamic_stack.go
```diff
@@ -0,0 +1,811 @@
...
+	for ret := range urlCheckResultChan {
+		if _, ok := urlRespCount[ret.url]; !ok {
+			urlRespCount[ret.url] = make(map[string]int)
+		}
+		payload := fmt.Sprintf("code: %v, body: %v, err: %v", ret.statusCode, ret.body, ret.err)
+		urlRespCount[ret.url][payload] += 1
+	}
+	cancel()
+	Expect(<-errChan).To(BeNil())
+	return urlRespCount
```
jaypipes: Where are we asserting that a particular percentage of weighted traffic was routed to a specific backend?

test/e2e/fishapp/dynamic_stack.go
```diff
@@ -0,0 +1,811 @@
...
+//		-> /path2 -> vn5(50)
+//				  -> vn6(50)
+//	vs5 -> /path1 -> vn7(50)
+//				  -> vn8(50)
+//		-> /path2 -> vn9(50)
+//				  -> vn10(50)
+// =======virtual nodes =========
+//  vn1 -> vs1,vs2,vs3,vs4
+//  vn2 -> vs5,vs1,vs2,vs3
+//  vn3 -> vs4,vs5,vs1,vs2
```
jaypipes: I have some concerns with structuring test fixtures in this way.

First, the mesh does not represent anything corresponding to a real-world use case. Because of this, it is difficult for readers of the test code to reason about. Having a real-world scenario described in a test fixture enables readers to reasonably understand and explain test assertions.

Secondly, you are always creating a symmetrical mesh. In other words, the "dynamic" stack is designed so that it actually *always* produces an mesh where each virtual service is constructed in an identical manner. This will inevitably lead to a greater chance of false positive test results that tend to exclude edge/corner cases that materialize when you have scenarios that are "uneven" (i.e. more real-worldish ;) ).

As an example of how this works out, let me take this assertion function example from this patch:

```go
// expects connectivity and routing works correctly
func (s *DynamicStack) ExpectBehavesCorrectly(ctx context.Context, f *framework.Framework) {
	time.Sleep(3 * time.Minute)
	vsByName := make(map[string]*appmeshv1beta1.VirtualService)
	for i := 0; i != s.VirtualServicesCount; i++ {
		vs := s.createdServiceVSs[i]
		vsByName[vs.Name] = vs
	}

	podsConnectivityCheckResult := make(map[string]map[string]map[string]int)
	for i := 0; i != s.VirtualNodesCount; i++ {
		dp := s.createdNodeDPs[i]
		vn := s.createdNodeVNs[i]

		var vsList []*appmeshv1beta1.VirtualService
		for _, backend := range vn.Spec.Backends {
			vsList = append(vsList, vsByName[backend.VirtualService.VirtualServiceName])
		}
		ret := s.checkDeploymentVirtualServiceConnectivity(ctx, f, dp, vsList)
		for k, v := range ret {
			podsConnectivityCheckResult[k] = v
		}
	}
	prettyJSON, err := json.MarshalIndent(podsConnectivityCheckResult, "", "    ")
	Expect(err).NotTo(HaveOccurred())
	utils.Logf("%v", string(prettyJSON))
}
```

Now, when I read the above, I don't get any semantic or scenario-specific context at all. I don't really know **why** a particular virtual node or virtual service is expected to exist or not exist. It's just too abstracted from reality.

If instead of defining a generic always-symmetric mesh with abstract node and service names, imagine creating **named** real-world mesh examples with realistic names:

```go
type MeshFixture interface {
    Name() string
    Deploy() error
    Check() error
    Cleanup() error
}

// A blue-green deployment scenario, implements MeshFixture
type BlueGreenScenario struct {
    webapp *corev1.Deployment
    service *corev1.Service
    virtService *appmeshv1beta1.VirtualService
    routeConfig shared.RouteToWeightedVirtualNodes
    blueNode *appmeshv1beta1.VirtualNode
    greenNode *appmeshv1beta1.VirtualNode
}

func (d *BlueGreenScenario) Deploy() error {
    // Create the Deployment, Service, VirtualService, Mesh and VirtualNode objects
    // NOTE: do NOT create the VirtualRouter yet
}

func (d *BlueGreenScenario) Check() error {
    // Assert that the Mesh, VirtualNodes, VirtualService, Deployment and Service objects exist properly
    ...
    // Create the VirtualRouter which establishes the weighted routing of traffic to
    // the Green virtual node
    ...
    // Check that traffic is flowing to the Green virtual node now
   ...
}
```

The individual assertions within the `Check()` method will, by nature of this setup, be much more descriptive. Instead of failure messages looking like this:

```go
	Expect(len(pods.Items)).NotTo(BeZero())
```
```
check connectivity of pod 1-a2ka to service 1
Expected 0 to not be 0
```

to instead be something like this:

```
	Expect(len(d.webapp.Pods)).NotTo(BeZero())
```
```
        checking connectivity of pod green-a2ka to service web-app
```

test/e2e/fishapp/dynamic_stack.go
```diff
@@ -0,0 +1,811 @@
...
+				return false
+			})
+			if cmNamespaceID == "" {
+				err := errors.Errorf("cannot find cloudMap namespace with name %s", s.cloudMapNamespace)
+				utils.Logf("failed to delete cloudMap namespace: %s due to %v", s.cloudMapNamespace, err)
+				deletionErrors = append(deletionErrors, err)
+				return
+			}
+
+			// hummm, let's fix the controller bug in test cases first xD:
```
jaypipes: Can you link to the controller bug you are fixing here please?

test/e2e/fishapp/dynamic_stack_test.go
```diff
@@ -0,0 +1,66 @@
...
+				BackendsCountPerVirtualNode: 2,
+				ReplicasPerVirtualNode:      3,
+				ConnectivityCheckPerURL:     100,
+			}
+		})
+		It("should behaves correctly", func() {
+			dynamicStack.ExpectBehavesCorrectly(ctx, f)
+			var ans int
+			fmt.Println("press any thing to continue cleanup")
+			_, _ = fmt.Scan(&ans)
```
jaypipes: Is there a reason to add this manual step here?

test/e2e/fishapp/dynamic_stack.go
```diff
@@ -0,0 +1,811 @@
...
+	}
+	if s.ServiceDiscoveryType == shared.CloudMapServiceDiscovery {
+		if errs := s.deleteCloudMapNamespace(ctx, f); len(errs) != 0 {
+			deletionErrors = append(deletionErrors, errs...)
+		}
+	}
+	if errs := s.deleteMeshAndNamespace(ctx, f); len(errs) != 0 {
+		deletionErrors = append(deletionErrors, errs...)
+	}
+	Expect(len(deletionErrors)).To(BeZero())
```
bcelenza: What happens if the cleanup step fails? Do resources need to be manually removed, or can you retry cleanup only?

test/e2e/fishapp/dynamic_stack_test.go
```diff
@@ -0,0 +1,66 @@
...
+		BeforeEach(func() {
+			dynamicStack = &fishapp.DynamicStack{
+				ServiceDiscoveryType:        shared.CloudMapServiceDiscovery,
+				VirtualServicesCount:        5,
+				VirtualNodesCount:           5,
+				RoutesCountPerVirtualRouter: 2,
+				TargetsCountPerRoute:        4,
+				BackendsCountPerVirtualNode: 2,
+				ReplicasPerVirtualNode:      3,
+				ConnectivityCheckPerURL:     100,
```
bcelenza: Non-blocking: I think it would be nice to be able to define these via a template or other argument format so they can be modified quickly and the tests run with different configurations.

test/e2e/framework/utils/log.go
```diff
@@ -0,0 +1,25 @@
...
+}
+
+func Logf(format string, args ...interface{}) {
+	log("INFO", format, args...)
+}
+
+func Failf(format string, args ...interface{}) {
+	msg := fmt.Sprintf(format, args...)
+	ginkgo.Fail(msg)
+}
```
bcelenza: Possible to just use something like logrus to eliminate needing this?

test/e2e/fishapp/dynamic_stack.go
M00nF1sh: Added a test based on chi-square under pvalue 0.001 :D

test/e2e/fishapp/dynamic_stack.go
M00nF1sh: Hi Jay, I agree this dynamic test case is not enough.

The the whole point of this PR is to create a test framework and a test case to cover the basic functionalities of AppMesh(multiple service/multiple node/multiple route/multiple target/multiple backend). **It's not intended to cover all E2E scenarios given the timeline i have.**
The dynamic stack is designed in a way to make it reusable for scale tests as well, which allows to do scale test with any dimensions.

There are more tests cases should be covered, e.g. add/remove route, change weights, etc.

The test framework is structured in a way to add new test cases easily. 
e.g. 
* the folder *test/e2e/fishapp* is specific to this `colorApp+httpProxy` container, it can be used to test any http-based appMesh scenarios. 
* the file *test/e2e/fishapp/dynamic_stack is specific to this symmetric mesh based on "fishapp".
* [**extension point**] more tests based on this "fishapp" container(which can be used as either client/server or both at same time)
* [**extension point**] other test scenario not rely on "fishapp" can be added like `test/e2e/color-app`, `test/e2e/bluegreen-app`, etc...

BTW, i didn't design a MeshFixture like interface because these tests tends to be highly depends on use case, some scenario based tests may have deploy & test steps interleaved, there is no value added to have such interface.

test/e2e/fishapp/dynamic_stack.go
M00nF1sh: thanks, added the bug links into comments :D 
```
// https://github.com/aws/aws-app-mesh-controller-for-k8s/issues/107
// https://github.com/aws/aws-app-mesh-controller-for-k8s/issues/131
```

test/e2e/fishapp/dynamic_stack_test.go
M00nF1sh: nice catch, it's used for my debug steps to hold the test cleanup for trouble-shotting, removed :D

test/e2e/fishapp/dynamic_stack.go
M00nF1sh: the cleanup steps are independent from each other(failed steps will report errors).
And yes, if resources are left over, we need to manually removed. I don't think add retry helps before we saw resource leaks scenarios. (like apiserver crash etc).
We'll monitor test failures and use dedicated account to run tests. If we saw resource leaks, we'll add handling for them correspondingly.

test/e2e/fishapp/dynamic_stack_test.go
M00nF1sh: Thanks, i think it's a valuable idea. 
However, since tests is highly dynamic, and i think code itself is better than template :D. (e.g. we can have multiple tests tests different dimension combinations specified, currently i only specified one)

test/e2e/framework/utils/log.go
M00nF1sh: humm, do we need to do structured logging in test cases? this can be an improvement :D

bcelenza: Structured as in JSON: no probably not, but the default format is highly similar to what you've defined here.

test/e2e/framework/resource/deployment/manager.go
```diff
@@ -0,0 +1,61 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+)
+
+type Manager interface {
+	WaitUntilDeploymentReady(ctx context.Context, dp *appsv1.Deployment) (*appsv1.Deployment, error)
+
```
nckturner: nit: remove whitespace

test/e2e/framework/k8s/port_forward.go
nckturner: I think what Kiran meant is to add a little of the explanation to the comment above--which I think is a good idea.

test/e2e/fishapp/dynamic_stack_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+		AfterEach(func() {
+			for _, stack := range stacksPendingCleanUp {
+				stack.Cleanup(ctx, f)
+			}
+		})
+
+		for _, sdType := range []shared.ServiceDiscoveryType{shared.DNSServiceDiscovery, shared.CloudMapServiceDiscovery} {
+			func(sdType shared.ServiceDiscoveryType) {
+				It(fmt.Sprintf("should behaves correctly with service discovery type %v", sdType), func() {
```
nckturner: "should behave correctly" is a bit vague, can you 1. split this into multiple tests and 2. improve the description to something like "should helm install and upgrade for controller and injector".

M00nF1sh: @nckturner what do you mean by split this into multiple tests. 
There are two test currently DNS and CloudMap. and within the test unit, the tests are sequential, so not really split able.
`install old controller`  => `deploy stack` => `verify behavior` => 
`install new controller` => `verify behavior against old stack` =>
`deploy new stack` => `verify behavior against new stack`

Also, for the behavior, the test is just check connectivity and result distribution.

test/e2e/framework/resource/deployment/manager.go
M00nF1sh: :D addressed

test/e2e/framework/utils/log.go
M00nF1sh: switched to use zap xD

.circleci/config.yml
```diff
@@ -1,66 +1,122 @@
...
+    description: "checkout code and cache go dependencies"
+    steps:
+      - checkout
+      - restore_cache:
+          key: go-mod-{{ checksum "go.sum" }}
+      - run:
+          name: go mod download
+          command: go mod download
+      - save_cache:
+          key: go-mod-{{ checksum "go.sum" }}
```
jaypipes: nice work above with the go mod caching. :+1:

pkg/client/clientset/versioned/clientset.go
```diff
@@ -47,9 +49,14 @@ func (c *Clientset) Discovery() discovery.DiscoveryInterface {
...
 // NewForConfig creates a new Clientset for the given config.
+// If config's RateLimiter is not set and QPS and Burst are acceptable,
+// NewForConfig will generate a rate-limiter in configShallowCopy.
 func NewForConfig(c *rest.Config) (*Clientset, error) {
 	configShallowCopy := *c
 	if configShallowCopy.RateLimiter == nil && configShallowCopy.QPS > 0 {
+		if configShallowCopy.Burst <= 0 {
+			return nil, fmt.Errorf("Burst is required to be greater than 0 when RateLimiter is not set and QPS is set to greater than 0")
+		}
 		configShallowCopy.RateLimiter = flowcontrol.NewTokenBucketRateLimiter(configShallowCopy.QPS, configShallowCopy.Burst)
```
jaypipes: Note that you could have put these changes into an entirely separate PR/commit, since the above are not directly related to the purpose of this PR.

scripts/ci_e2e_test.sh
```diff
@@ -0,0 +1,154 @@
...
+# Arguments:
+#   image_repo
+#   image_tag
+#   image_name
+#
+#######################################
+build_push_controller_image() {
+    declare -r image_repo="$1" image_tag="$2" image_name="$3"
+
+    ecr::ensure_repository "${image_repo}"
```
jaypipes: hmm, the above looks familiar :)

scripts/ci_e2e_test.sh
```diff
@@ -0,0 +1,154 @@
...
+      return 1
+  fi
+  build_push_controller_image "${IMAGE_REPO_APP_MESH_CONTROLLER}" "${IMAGE_TAG}" "${image_name}"
+
+  echo $image_name
+  trap "cleanup_cluster" EXIT
+  setup_cluster
+  test_controller_image "${image_name}"
+}
+
```
jaypipes: Good stuff in this script file. Nice and clean. ++

test/e2e/fishapp/dynamic_stack.go
jaypipes: > But the purpose of this PR is to create a test framework and a test case to cover the basic functionalities of AppMesh(multiple service/multiple node/multiple route/multiple target/multiple backend). **It's not intended to cover all E2E scenarios given the timeline i have.**

I didn't say that I expected you to cover all possible e2e scenarios :)

> The dynamic stack is designed in a way to make it reusable for scale tests as well, which allows to do scale test with configurable dimensions. (I found an appmesh dataplane bug with it )

Yes, I understand that, and the dynamic stack fixture you've created here is a good way to do fuzz, edge/negative, and stress testing. No problem with that.

What I would like to see in a followup PR is an approach that covers the standard use cases/scenarios that AppMesh is designed to cover (blue/green, A/B, etc) in a manner that is easier for readers of the test cases to figure out expectations. Again, it's cool what you've done here. I just think a future enhancement to simplify the scenarios would be useful.

> There are more tests cases yet to be covered, e.g. add/remove route, change weights, http2, grpc, tls, retry, etc.

Yes, agreed completely.

> The test framework is structured in a way to add new test cases easily.
> e.g.
> 
> * the folder _test/e2e/fishapp_ is specific to this `colorApp+httpProxy` container, it can be used to test any http-based appMesh scenarios.
> * the file *test/e2e/fishapp/dynamic_stack is specific to this symmetric mesh based on "fishapp".
> * [**extension point**] more tests based on this "fishapp" container(which can be used as either client/server or both at same time), e.g. `test/e2e/fishapp/bluegreen_stack`
> * [**extension point**] other test scenario not rely on "fishapp" can be added like `test/e2e/color-app`, `test/e2e/bluegreen-app`, etc...

Yep, no disagreement on this.

> BTW, i didn't design a MeshFixture like interface because these tests tends to be highly depends on use case, some scenario based tests may have deploy & test steps interleaved, there is no value added to have such interface.

My suggestion is always code to an interface, not a concrete struct. Makes things more extensible and easier to reason about in the long run. :)

.circleci/config.yml
M00nF1sh: 😄

pkg/client/clientset/versioned/clientset.go
M00nF1sh: nice catch :D. i reverted the version change to controller-gen tools 😄

scripts/ci_e2e_test.sh
M00nF1sh: haha, i steal this ensure repo from CNI repo :D

scripts/ci_e2e_test.sh
M00nF1sh: 😃

test/e2e/fishapp/dynamic_stack.go
M00nF1sh: 😄 TBH, my original version of E2E test is using a blue green deployment color-app. But it cannot cover the case of multiple-path and multiple virtual service. So instead I took this dynamic symmetrical mesh approach.  (I was also asked to script my scale tests, so it can be reused xD).

Yeah, we can add more real-word e2e tests into this in the future :D

PR #152, opened 2020-04-02
add CRD API types for v1beta2

M00nF1sh:
Adding new CRD API version for v1beta2(based on designs https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/150)

## Note for reviewer:
The PR is separated into three commit, only the 2nd one needs to be reviewed while other two is auto-generated.
1. "auto-generate yaml for CRDs" is auto-generated by
    1. `kubebuilder edit --multigroup=true`
    2. `kubebuilder create api --group appmesh --version v1beta2 --kind Mesh --namespaced=false`
    3. `kubebuilder create api --group appmesh --version v1beta2 --kind VirtualNode`
    4. `kubebuilder create api --group appmesh --version v1beta2 --kind VirtualRouter`
    5. `kubebuilder create api --group appmesh --version v1beta2 --kind VirtualService`
2. **"add Fields into CRD APIs" contains real manual changes**
3. "add CRD API types for v1beta2" is auto-generated by
    1. `make manifests`
## changes in addition to v1beta1:
   1. all changes mentioned in design: https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/150
   2. ServiceDiscoveryType is removed from Mesh.spec. (since it's not used anywhere and not presents in AppMeshAPI).
   3. DNSServiceDiscovery.hostName renamed to be "DNSServiceDiscovery.hostname". (to match appMesh API and k8s api convention)
   4. ServiceDiscovery.cloudMap renamed to be "ServiceDiscovery.awsCloudMap" (to align with appMesh API)
   5. 'VirtualRouter.spec.routes.[].http' to be 'VirtualRouter.spec.routes.[].httpRoute'. (to align with appMesh API, same change is done for `tcp/https/grpc`)
   6. 'perRetryTimeoutMillis' -> perRetryTimeout' with a defined Duration struct. (to align with appMesh API)
   7 .'virtualNode.spec.serviceDiscovery.awsCloudMap.attributes' from `map[string]string` to be array of awsCloudMapAttribute. (to align with appMesh API)
   8. [meshCondition](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/apis/appmesh/v1beta1/types.go#L48) renamed to `conditions`
   9. multiple golang field rename to align with appmesh and APIConvention.(the crd json field don't change)


apis/appmesh/v1beta2/virtualrouter_types.go
```diff
@@ -0,0 +1,367 @@
...
+type GRPCRetryEvent string
+
+// HttpRetryPolicy refers to https://docs.aws.amazon.com/app-mesh/latest/APIReference/API_HttpRetryPolicy.html
+type HttpRetryPolicy struct {
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=25
+	// +optional
+	HTTPRetryEvents []HTTPRetryEvent `json:"httpRetryEvents,omitempty"`
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=1
```
jaypipes, line 132: Not entirely sure what an enforced single-item array gives us that wouldn't be better represented as a scalar element, but OK.

apis/appmesh/v1beta2/types.go
```diff
@@ -0,0 +1,194 @@
...
+)
+
+// PortMapping refers to https://docs.aws.amazon.com/app-mesh/latest/APIReference/API_PortMapping.html
+type PortMapping struct {
+	// The port used for the port mapping.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=65535
+	Port int64 `json:"port"`
+	// The protocol used for the port mapping.
+	Protocol Protocol `json:"protocol"`
```
achevuru, line 20: Can we add validation to this field as well to make sure we don't allow some random string?

achevuru: I see that 'Protocol' is already tagged with validation. Ignore my previous comment.

apis/appmesh/v1beta2/virtualservice_types.go
```diff
@@ -17,25 +17,79 @@ limitations under the License.
...
+	// The reason for the condition's last transition.
+	// +optional
+	Reason *string `json:"reason,omitempty"`
+	// A human readable message indicating details about the transition.
+	// +optional
+	Message *string `json:"message,omitempty"`
+}
 
 // VirtualServiceSpec defines the desired state of VirtualService
 type VirtualServiceSpec struct {
```
achevuru, line 70: Don't we need an 'awsName' field for VirtualService as well since we create an AWS App Mesh object for this as well?

apis/appmesh/v1beta2/virtualrouter_types.go
M00nF1sh: Hi jay, i got it from the APIDatastructure: https://docs.aws.amazon.com/app-mesh/latest/APIReference/API_HttpRetryPolicy.html
I think it's good to align with api(which uses a single-item array), since it will help us align with api changes in the future. (e.g. more tcp events)

M00nF1sh:
> I see in the second commit that you "manually" added the Fields for each of the Resource and normal struct types. Is that correct? Or did you generate the `*types.go` from some other schema document?

I added these validations manually given the description from doc, it should be possible to auto-generate these from AWSSDK's JSON manifest :D.
BTW, do you think we should keep these changes in our k8s(which varies from appMesh's data structure) (these changes exists in v1beta1, and personally i think there are good :D)
1. virtualNode.spec.serviceDiscovery.awsCloudMap -> cloudMap
1. virtualRouter.spec.routes.[].httpRoute -> 'http' (similar change for http2/grpc/tcp)
1. 'perRetryTimeout'->'perRetryTimeoutMillis'
1. 'virtualNode.spec.serviceDiscovery.awsCloudMap.attributes' -> map[string]string.

apis/appmesh/v1beta2/types.go
M00nF1sh: :D yeah. the annotation on types works :D

apis/appmesh/v1beta2/virtualservice_types.go
M00nF1sh: it's will be the hosts. an k8s virtualService corresponding to multiple appmesh virtualService.
each host will become a appMesh virtualService's awsName :D

achevuru: Ok, so are we going to create multiple Virtual Service App Mesh objects based on hosts array for a single K8S Virtual Service? Let's say 1 for "svc-1.my-ns.svc.cluster.local" and 1 for "svc-1" and so on...?

M00nF1sh: yes :D.
and by default, the host array will only have one entry, which is the VirtualService CRD's name.

M00nF1sh: FYI, i changed it to just be 'awsName' instead of 'hosts' array, per https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/150#discussion_r404365334

apis/appmesh/v1beta2/mesh_types.go
```diff
@@ -17,25 +17,73 @@ limitations under the License.
...
 // MeshSpec defines the desired state of Mesh
+// refers to https://docs.aws.amazon.com/app-mesh/latest/APIReference/API_MeshSpec.html
 type MeshSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
-
-	// Foo is an example field of Mesh. Edit Mesh_types.go to remove/update
-	Foo string `json:"foo,omitempty"`
+	// AWSName is the AppMesh Mesh object's name.
+	// If unspecified, it defaults to be "${name}" of k8s Mesh
```
efe-selcuk, line 67: What is the "name of the k8s Mesh"? Doesn't this get mapped from the CRDs?

M00nF1sh: it's the name of the mesh CR(custom resource) created in k8s.

efe-selcuk:
Based off of the discussions above and #150 LGTM

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,7 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: m00nf1sh/misc
```
achevuru, line 7: Might want to change the newName field here?

M00nF1sh: sorry, these changes shouldn't be there, it's for follow up changes after this pr merged. I miss pushed to this branch. removed :D

M00nF1sh:
Talked with jay over chime.  we'll favor golang's best practice over sdk ones.
So like "meshArn" will be "meshARN". I'll submit a follow up PR to change all of them :D

PR #157, opened 2020-04-10
add webhooks for designate mesh and defaulting fields

M00nF1sh:
# Changes
## Note for reviewer:
1. This PR is separated to two commit, while "auto-generate yaml and code for CRD APIs" is auto-generated.
2. Files within /mocks is autogenerated by `scripts/gen_mocks.sh`
## code structure
1. /pkg for shared utility code.
1. /webhooks for webhooks
1. /mocks for generated mock files
## build scripts changes
1. added hack/gen_mocks.sh to generate mocks for interfaces.
1. modified `config/default/kustomization.yaml` to 
    * added support for webhooks
    * changed to use "appmesh-" as component prefix.
1. Dockerfile change to support new folders
## add webhooks for designate mesh and defaulting fields
1. The mutating webhooks for CRDs is responsible for designate and defaulting fields for objects
2. The validating webhook for CRDs is responsible for validating fields, currently only awsName's immutability is enforced.
3. The mutating webhook will act as Injector. Currently just a fake implementation that add a annotation to denote virtualNode.
4. I added separate `webhook.Mutator` and `webhook.Validator` instead of reuse the ones in controller-runtime: [defaulter](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/admission/defaulter.go), [validator](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/admission/validator.go) due to below reasons:
   * our mutating webhook for designate membership should be able to error out
   * our mutating and validating requires more info(like meshDesignator k8sClient)  to be injected instead of just the object itself.
   
   Note: the object type function signature is `Prototype(req admission.Request) (runtime.Object, error)` instead of just `Prototype() runtime.Object` is to make it flexible, so that a dynamic webhook for multiple webhook is possible. (for our use case, `Prototype() runtime.Object` is enough since each webhook only handles a single CRD)

## Steps to test
1. Register an dockerHub or ECR repository and setup credentials. e.g. m00nf1sh/misc
2.  
    ```bash
    export IMG=m00nf1sh/misc:appmesh_dev
    ```
3.  
    ```bash
    make docker-build && make docker-push
    ```
4. Prepare your cluster's kubeconfig, you should be able to run "kubectl" commands
5. Install certificate manager to your cluster by https://cert-manager.io/docs/installation/kubernetes/ (we'll remove dependency on cert-manager in later PRs)
6. Cleanup previously installed v1beta1 CRD if any. (we'll come up with changes without require to do this in future PRs)
   ```
   kubectl delete -f https://raw.githubusercontent.com/aws/eks-charts/master/stable/appmesh-controller/crds/crds.yaml
   ```
7. Deploy controller to cluster.
    ```bash
    make deploy
    ```


pkg/mesh/membership_designator.go
```diff
@@ -0,0 +1,64 @@
...
+		return nil, errors.Errorf("failed to find matching mesh for namespace: %s, expecting 1 but found %d",
+			obj.GetNamespace(), 0)
+	}
+	if len(meshCandidates) > 1 {
+		var meshCandidatesNames []string
+		for _, mesh := range meshCandidates {
+			meshCandidatesNames = append(meshCandidatesNames, mesh.Name)
+		}
+		return nil, errors.Errorf("failed to find matching mesh for namespace: %s, expecting 1 but found %d: %s",
+			obj.GetNamespace(), len(meshCandidates), strings.Join(meshCandidatesNames, ","))
```
achevuru, line 61: Might be clear if we instead say "Found multiple matching meshes ..." ?

webhooks/appmesh/mesh_validator.go
```diff
@@ -0,0 +1,64 @@
...
+
+var _ webhook.Validator = &meshValidator{}
+
+type meshValidator struct {
+}
+
+func (v *meshValidator) Prototype(req admission.Request) (runtime.Object, error) {
+	return &appmesh.Mesh{}, nil
+}
+
```
achevuru, line 30: Should we block a meshSpec selecting on system namespace ..say "kube-system", in their mesh spec as part of this?

pkg/mesh/membership_designator.go
M00nF1sh: updated :D thanks

webhooks/appmesh/mesh_validator.go
M00nF1sh: i don't think we should do that within code. customer should be have a controller-level flag to controller that if we decide to expose such functionality :D

PR #163, opened 2020-04-17
add default throttling and supports override via --aws-api-throttle

M00nF1sh:
Add default throttling for aws apis.
customer can override it via `--aws-api-throttle`.
the default value can be viewed:
```
      --aws-api-throttle serviceOperationsThrottleConfig   throttle settings for aws APIs, format: serviceID1:operationRegex1=rate:burst,serviceID2:operationRegex2=rate:burst (default App Mesh:^Describe|List=40:5,App Mesh:^Create|Update|Delete=8:5,ServiceDiscovery:^ListNamespaces=1:8,ServiceDiscovery:^ListServices=1:8,ServiceDiscovery:^CreateService=8:80,ServiceDiscovery:^ListInstances=40:400,ServiceDiscovery:^RegisterInstance|DeregisterInstance=4:80
```

### DIFF
--- a/pkg/aws/throttle/config.go
+++ b/pkg/aws/throttle/config.go
@@ -1,0 +1,102 @@
+package throttle
+
+import (
+	"fmt"
+	"github.com/pkg/errors"
+	"github.com/spf13/pflag"
+	"golang.org/x/time/rate"
+	"regexp"
+	"sort"
+	"strconv"
+	"strings"
+)
+
+type throttleConfig struct {
+	operationPtn *regexp.Regexp
+	r            rate.Limit
+	burst        int
+}
+
+var _ pflag.Value = &ServiceOperationsThrottleConfig{}
+
+// serviceOperationsThrottleConfig is throttleConfig for each service's operations.
+// It supports to be configured using flags with format like "${serviceID}:${operationRegex}={rate}:{burst}"
+// e.g. "appmesh:DescribeMesh=1.3:5,appmesh:Create.*=1.7:3"
+// Note: default throttle for each service will be cleared if any override is set for that service.
+type ServiceOperationsThrottleConfig struct {
+	// service:operationRegex:config
+	value map[string][]throttleConfig
+}
+
+func (c *ServiceOperationsThrottleConfig) String() string {
+	if c == nil {
+		return ""
+	}
+
+	var configs []string
+	var serviceIDs []string
+	for serviceID := range c.value {
+		serviceIDs = append(serviceIDs, serviceID)
+	}
+	sort.Strings(serviceIDs)
+	for _, serviceID := range serviceIDs {
+		for _, operationsThrottleConfig := range c.value[serviceID] {
+			configs = append(configs, fmt.Sprintf("%s:%s=%v:%d",
+				serviceID,
+				operationsThrottleConfig.operationPtn.String(),
+				operationsThrottleConfig.r,
+				operationsThrottleConfig.burst,
+			))
+		}
+	}
+	return strings.Join(configs, ",")
+}
+
+func (c *ServiceOperationsThrottleConfig) Set(val string) error {
+	valueOverride := make(map[string][]throttleConfig)
+	configPairs := strings.Split(val, ",")
+	for _, pair := range configPairs {
+		kv := strings.Split(pair, "=")
+		if len(kv) != 2 {
+			return errors.Errorf("%s must be formatted as serviceID:operationRegex=rate:burst", pair)
+		}
+		serviceIDOperationRegexPair := strings.Split(kv[0], ":")
+		if len(serviceIDOperationRegexPair) != 2 {
+			return errors.Errorf("%s must be formatted as serviceID:operationRegex", kv[0])
+		}
+		rateBurstPair := strings.Split(kv[1], ":")
+		if len(rateBurstPair) != 2 {
+			return errors.Errorf("%s must be formatted as rate:burst", kv[1])
+		}
+		serviceID := serviceIDOperationRegexPair[0]
+		operationPtn, err := regexp.Compile(serviceIDOperationRegexPair[1])
+		if err != nil {
+			return errors.Errorf("%s must be valid regex expression for operation", serviceIDOperationRegexPair[1])
+		}
+		r, err := strconv.ParseFloat(rateBurstPair[0], 64)
+		if err != nil {
+			return errors.Errorf("%s must be valid float number as rate for operations per second", rateBurstPair[0])
+		}
+		burst, err := strconv.Atoi(rateBurstPair[1])
+		if err != nil {
+			return errors.Errorf("%s must be valid integer as burst for operations", rateBurstPair[1])
+		}
+		valueOverride[serviceID] = append(valueOverride[serviceID], throttleConfig{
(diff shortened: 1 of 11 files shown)


pkg/aws/throttle/defaults.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+// NewDefaultServiceOperationsThrottleConfig returns a ServiceOperationsThrottleConfig with default settings.
+func NewDefaultServiceOperationsThrottleConfig() *ServiceOperationsThrottleConfig {
+	return &ServiceOperationsThrottleConfig{
+		value: map[string][]throttleConfig{
+			appmesh.ServiceID: {
+				{
+					operationPtn: regexp.MustCompile("^Describe|List"),
+					r:            rate.Limit(40),
```
achevuru, line 17: How about converting these values in to consts with descriptive names?

pkg/aws/throttle/config.go
```diff
@@ -0,0 +1,102 @@
...
+	for _, pair := range configPairs {
+		kv := strings.Split(pair, "=")
+		if len(kv) != 2 {
+			return errors.Errorf("%s must be formatted as serviceID:operationRegex=rate:burst", pair)
+		}
+		serviceIDOperationRegexPair := strings.Split(kv[0], ":")
+		if len(serviceIDOperationRegexPair) != 2 {
+			return errors.Errorf("%s must be formatted as serviceID:operationRegex", kv[0])
+		}
+		rateBurstPair := strings.Split(kv[1], ":")
```
achevuru, line 67: Would it be good to range limit the values users can choose for the rate/burst? Maybe good to prevent them from choosing unreasonable values on both ends (high and low)

pkg/aws/throttle/predict.go
```diff
@@ -0,0 +1,32 @@
...
+func matchServiceOperation(serviceID string, operation string) Predict {
+	return func(r *request.Request) bool {
+		if r.Operation == nil {
+			return false
+		}
+		return r.ClientInfo.ServiceID == serviceID && r.Operation.Name == operation
+	}
+}
+
+func matchServiceOperationPattern(serviceID string, operationPtn *regexp.Regexp) Predict {
```
efe-selcuk, line 25: What does "ptn" stand for?

efe-selcuk: Oh, pattern. I don't think we're typing it often enough to shorten it :P

pkg/aws/throttle/defaults.go
M00nF1sh: tbh, personally I think it's more clear to directly use them instead of using constants.
since all of these literals is used only once(there are multiple "1" or "8", but it's different per operationPtn). 
given they are only used once and we have many configuration per service/operation, it's clear to just list them in structs

pkg/aws/throttle/predict.go
```diff
@@ -0,0 +1,32 @@
+package throttle
+
+import (
+	"github.com/aws/aws-sdk-go/aws/request"
+	"regexp"
+)
+
+type Predict func(r *request.Request) bool
```
efe-selcuk, line 8: What is this "predicting"? Seems like it's a matcher of sorts?

pkg/aws/throttle/config.go
M00nF1sh: I don't think there are sensible high and low. (except they must be non-negative).
These are advanced configurations, and ppl should only override them if they got limit increase from aws side. (where they can got corresponding rate & burst config from aws)

pkg/aws/throttle/throttler_test.go
```diff
@@ -0,0 +1,199 @@
...
+					t.Errorf("count = %d, want %d (ideal %f", count, want, ideal)
+				}
+				// We should get very close to the number of requests allowed.
+				if want := int64(ideal - 1); count < want {
+					t.Errorf("count = %d, want %d (ideal %f", count, want, ideal)
+				}
+			},
+		},
+		{
+			name: "two limiter, both matching",
```
efe-selcuk, line 134: This test is ensuring that the burst is honored? Shouldn't the two limits be different so we can check which one is honored? Or is that non-deterministic?

pkg/aws/throttle/throttler_test.go
```diff
@@ -0,0 +1,199 @@
...
+)
+
+func Test_throttler_InjectHandlers(t *testing.T) {
+	throttler := &throttler{}
+	handlers := request.Handlers{}
+	throttler.InjectHandlers(&handlers)
+	assert.Equal(t, 1, handlers.Sign.Len())
+}
+
+func Test_throttler_beforeSign(t *testing.T) {
```
efe-selcuk, line 22: Can we describe what this is verifying either as a comment or more descriptive test cases? Clearly I had questions :P

pkg/aws/throttle/throttler_test.go
```diff
@@ -0,0 +1,199 @@
...
+				ideal := 5 + 1*elapsedDuration.Seconds()
+				// We should never get more requests than allowed.
+				if want := int64(ideal + 1); count > want {
+					t.Errorf("count = %d, want %d (ideal %f", count, want, ideal)
+				}
+				// We should get very close to the number of requests allowed.
+				if want := int64(ideal - 1); count < want {
+					t.Errorf("count = %d, want %d (ideal %f", count, want, ideal)
+				}
+			},
```
efe-selcuk, line 167: Is this expecting at least 4 and at most 6? Am I reading it right?

pkg/aws/throttle/predict.go
M00nF1sh: I can also name it as "condition" is it ok to you?

pkg/aws/throttle/throttler_test.go
M00nF1sh: they have different rate :D. 
the purpose is to  just to check if there are different limits(rate+burst), the lower one get honered.

pkg/aws/throttle/throttler_test.go
M00nF1sh: yes

pkg/aws/throttle/throttler_test.go
efe-selcuk: Oh, I see, the fact that the lower one is 1 TPS means we'll get the burst of 5 and nothing more within 1 second. Ok cool.

pkg/aws/throttle/throttler_test.go
efe-selcuk: Why is it imprecise? Even with overhead from the code you should be able to burst 5 times within a second, right?

M00nF1sh: The purpose is to reduce false alarm due to CPU load when executing tests.

The whole purpose of the test is to make sure the limit is there and didn't fail within 10x. (that's why i choose 1 rate and 10 rate for this two limiter).
The test QPS is 100, so without throttle, it will be approaching 500 calls. ([4, 6] already precisely expressed our throttle purpose)

pkg/aws/throttle/throttler_test.go
M00nF1sh: added more descriptions :D

Issue #37, opened 2019-05-14
Distinguish Router port from Node port


dastbe:
Today, in the custom resource definition for a virtual service we define routes and a virtual router name, but do not support pass-through of router listeners. This makes it difficult for users to create a 1-1 correspondence between their k8s services/pods and their App Mesh VirtualService+VirtualRouter/VirtualNodes when the service port differs from the container port.

It would make sense to extend the spec of a control Virtual Service to something like 

```
...
      properties:
        spec:
          properties:
            meshName:
              type: string
            virtualRouter:
              type: object
              properties:
                name:
                  type: string
            listeners:
              type: array
              items:
                type: object
                properties:
                  portMapping:
                    properties:
                      port:
                        type: integer
                      protocol:
                        type: string
                        enum:
                          - tcp
                          - http
                          - grpc
                          - http2
                          - https
            routes:
...
```

which could then be passed through to VRouter create/update calls. It would also make sense in a new revision of the controller resource spec to make this field required, as users may find the behavior as exists surprising.

Some context on why/how things work as currently implemented.

To be backwards compatible with resources from the preview API that did not define router listeners, App Mesh allows listener-less virtual routers to operate in a "legacy" mode for http routes. Specifically, for each route we merge the listener ports of the targeted virtual nodes and use that as the router listener port for that route. This has some notable drawbacks

1. A user is unable to decouple changes between their node ports and their router port
1. If the target virtual nodes behind a route disagree on their listener port, App Mesh ends up dropping the route altogether

with a router listener port defined, disagreement between node ports is abstracted from customers, and an update to the node port doesn't impact the port clients use for making requests.

Additionally, TCP routes enforce the presence of the listener because they did not exist in the preview API.

nckturner:
Yeah, this is definitely needed!  Thanks for the issue!

kiranmeduri:
#46

PR #165, opened 2020-04-18
Add CRD for VirtualGateway APIs for v1beta2

rajal-amzn:
*Issue #, if available:*
[aws-app-mesh-roadmap #37](https://github.com/aws/aws-app-mesh-roadmap/issues/37)

Description of changes:
Adding VirtualGateway CRD API version for v1beta2(based on designs #150)

Note for reviewer:
The PR is separated into three commits, only the 2nd one needs to be reviewed while the other two are auto-generated.

"auto-generate changes for adding VirtualGateway API" is auto-generated by
"kubebuilder create api --group appmesh --version v1beta2 --kind VirtualGateway"
"Separate out common fields.." contains real manual changes where I have moved out the Virtual Node types to Virtual Node class and defined new ones for Virtual Gateway since we need flexibility to evolve both Virtual Node and Virtual Gateway separately
"Auto-generate CRD APIs for Virtual Gateway" is auto-generated by "make manifests && make generate"


apis/appmesh/v1beta2/virtualgateway_types.go
```diff
@@ -17,25 +17,262 @@ limitations under the License.
...
+const (
+	// VirtualGatewayActive is True when the AppMesh VirtualGateway has been created or found via the API
+	VirtualGatewayActive VirtualGatewayConditionType = "VirtualGatewayActive"
+)
+
+// +kubebuilder:validation:Enum=grpc;http;http2
+type VirtualGatewayPortProtocol string
+
+// VirtualGatewayPortMapping refers to https://docs.aws.amazon.com/app-mesh/latest/APIReference/API_PortMapping.html
+type VirtualGatewayPortMapping struct {
```
M00nF1sh, line 35: did we introduced different structure for these(e.g. healthCheckPolicy/PortMapping) in the SDK JSON?

The types here should be closely mapping to the types in SDK SON. (except use same case for Acronym, e.g. Tls->TLS). So that we can write tools to generate it in the future.

rajal-amzn: Yes. We have introduced separate structures for Virtual Gateway and Virtual Node in the AWS SDK. Refer [here](https://raw.githubusercontent.com/aws/aws-app-mesh-roadmap/master/appmesh-preview/service-model.json) for details.

apis/appmesh/v1beta2/types.go
```diff
@@ -226,3 +83,11 @@ type MeshReference struct {
 	// UID is the UID of Mesh CR
 	UID types.UID `json:"uid"`
 }
+
+// VirtualGatewayReference holds a reference to VirtualGateway.appmesh.k8s.aws
+type VirtualGatewayReference struct {
+	// Name is the name of Mesh CR
+	Name string `json:"name"`
+	// UID is the UID of Mesh CR
+	UID types.UID `json:"uid"`
```
karanvasnani, line 92: Shouldn't specifying `Namespace` here be sufficient as VirtualGateway is a Namespace-scoped resource unlike Mesh?

apis/appmesh/v1beta2/types.go
```diff
@@ -226,3 +83,11 @@ type MeshReference struct {
 	// UID is the UID of Mesh CR
 	UID types.UID `json:"uid"`
 }
+
+// VirtualGatewayReference holds a reference to VirtualGateway.appmesh.k8s.aws
+type VirtualGatewayReference struct {
+	// Name is the name of Mesh CR
```
karanvasnani, line 89: nit: VirtualGateway instead of Mesh here and below.

apis/appmesh/v1beta2/virtualgateway_types.go
```diff
@@ -17,25 +17,262 @@ limitations under the License.
...
+	// +kubebuilder:validation:Minimum=2000
+	// +kubebuilder:validation:Maximum=60000
+	// +optional
+	TimeoutMillis *int64 `json:"timeoutMillis,omitempty"`
+	// The number of consecutive failed health checks that must occur before declaring a virtual Gateway unhealthy.
+	// If unspecified, defaults to be 2
+	// +kubebuilder:validation:Minimum=2
+	// +kubebuilder:validation:Maximum=10
+	// +optional
+	UnhealthyThreshold *int64 `json:"unhealthyThreshold,omitempty"`
```
karanvasnani, line 95: From the model [here](https://raw.githubusercontent.com/aws/aws-app-mesh-roadmap/master/appmesh-preview/service-model.json) following should be required fields:
```
"HealthCheckPolicy": {
      "type": "structure",
      "required": [
        "healthyThreshold",
        "intervalMillis",
        "protocol",
        "timeoutMillis",
        "unhealthyThreshold"
      ]

```

rajal-amzn: These values have defaults if not specified in Virtual Node. Hence I followed the same convention.

apis/appmesh/v1beta2/types.go
rajal-amzn: Copy-paste error. :) Will fix it.

apis/appmesh/v1beta2/types.go
rajal-amzn: Yes thats right. This has to be namespace-scoped. Will change it.

M00nF1sh:
general rules: 
1. type/field always stick to SDK json with one exception, use same case for abbreviation like TLS/ACM/ARN
2. if a field is optional in sdk, make it optional in crd as well, and don't use mutating webhook to set default value.
3. if a field is required in sdk, but have sensible default in k8s context, we can make it optional, but use mutating webhook to set default value.
4. if a field is required in sdk, but don't have sensible default in k8s context, we make it required.

rajal-amzn:
Fixed the naming a

> some general rules: :D (so we need to change virtualNodeHealthCheckPolicy/tls/etc...)
> 
> 1. type/field always stick to SDK json with one exception, use same case for abbreviation like TLS/ACM/ARN
> 2. if a field is optional in sdk, make it optional in crd as well, and don't use mutating webhook to set default value.
> 3. if a field is required in sdk, but have sensible default in k8s context, we can make it optional, but use mutating webhook to set default value.
> 4. if a field is required in sdk, but don't have sensible default in k8s context, we make it required.

Thanks. I have fixed the naming in the latest revision. So, now the VirtualNode related APIs are just moved from types.go to virtualnode_types.go without any changes. Also validated the names of VirtualGateway types to match with the SDK.

PR #166, opened 2020-04-20
Consolidate the aws-app-mesh-inject repository

kishorj:
*Issue #, if available:*

*Description of changes:*
Consolidate the aws-app-mesh-inject repository

### DIFF
--- a/main.go
+++ b/main.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/mesh"
 	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/virtualnode"
+	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/appmeshinject"
 
 	appmeshv1beta2 "github.com/aws/aws-app-mesh-controller-for-k8s/apis/appmesh/v1beta2"
 	appmeshcontroller "github.com/aws/aws-app-mesh-controller-for-k8s/controllers/appmesh"
@@ -55,9 +56,13 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	var injectConfig appmeshinject.Config
+	injectConfig.BindFlags()
 	flag.Parse()
 
 	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
+
+	appmeshinject.New(injectConfig)
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:             scheme,
--- a/pkg/appmeshinject/config.go
+++ b/pkg/appmeshinject/config.go
@@ -1,0 +1,107 @@
+package appmeshinject
+
+import (
+	"errors"
+	"flag"
+)
+
+type Config struct {
+	// Injetion Settings
+	InjectDefault bool
+
+	// If enabled, an fsGroup: 1337 will be injected in the absence of it within pod securityContext
+	// see https://github.com/aws/amazon-eks-pod-identity-webhook/issues/8 for more details
+	EnableIAMForServiceAccounts bool
+
+	// Sidecar settings
+	SidecarImage  string
+	SidecarCpu    string
+	SidecarMemory string
+	//MeshName      string
+	Region    string
+	Preview   bool
+	LogLevel  string
+	EcrSecret bool
+
+	// Init container settings
+	InitImage          string
+	IgnoredIPs         string
+	EgressIgnoredPorts string
+
+	// Observability settings
+	InjectXraySidecar           bool
+	EnableStatsTags             bool
+	EnableStatsD                bool
+	InjectStatsDExporterSidecar bool
+	EnableJaegerTracing         bool
+	JaegerAddress               string
+	JaegerPort                  string
+	EnableDatadogTracing        bool
+	DatadogAddress              string
+	DatadogPort                 string
+}
+
+// MultipleTracer checks if more than one tracer is configured.
+func MultipleTracer(config *Config) bool {
+	j := config.EnableJaegerTracing
+	d := config.EnableDatadogTracing
+	x := config.InjectXraySidecar
+
+	return (j && d) || (d && x) || (j && x)
+}
+
+func (cfg *Config) BindFlags() {
+	flag.BoolVar(&cfg.InjectDefault, "inject-default", true,
+		`If enabled, sidecars will be injected in the absence of the corresponding pod annotation`)
+	flag.BoolVar(&cfg.EnableIAMForServiceAccounts, "enable-iam-for-service-accounts", true,
+		`If enabled, an fsGroup: 1337 will be injected in the absence of it within pod securityContext`)
+	flag.StringVar(&cfg.Region, "region" /*os.Getenv("APPMESH_REGION")*/, "",
+		"AWS App Mesh region")
+	flag.BoolVar(&cfg.Preview, "preview", false,
+		"Enable preview channel")
+	flag.StringVar(&cfg.LogLevel, "log-level" /*os.Getenv("APPMESH_LOG_LEVEL")*/, "info",
+		"AWS App Mesh envoy log level")
+	flag.BoolVar(&cfg.EcrSecret, "ecr-secret", false,
+		"Inject AWS app mesh pull secrets")
+	flag.StringVar(&cfg.SidecarImage, "sidecar-image", "840364872350.dkr.ecr.us-west-2.amazonaws.com/aws-appmesh-envoy:v1.12.3.0-prod",
+		"Envoy sidecar container image.")
+	flag.StringVar(&cfg.SidecarCpu, "sidecar-cpu-requests", "10m",
+		"Envoy sidecar CPU resources requests.")
+	flag.StringVar(&cfg.SidecarMemory, "sidecar-memory-requests", "32Mi",
+		"Envoy sidecar memory resources requests.")
+	flag.StringVar(&cfg.InitImage, "init-image", "111345817488.dkr.ecr.us-west-2.amazonaws.com/aws-appmesh-proxy-route-manager:v2",
+		"Init container image.")
+	flag.StringVar(&cfg.IgnoredIPs, "ignored-ips", "169.254.169.254",
+		"Init container ignored IPs.")
+	flag.BoolVar(&cfg.EnableJaegerTracing, "enable-jaeger-tracing", false,
+		"Enable Envoy Jaeger tracing")
+	flag.StringVar(&cfg.JaegerAddress, "jaeger-address", "appmesh-jaeger.appmesh-system",
+		"Jaeger address")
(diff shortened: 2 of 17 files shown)


lavignes:
Can the changes from aws/aws-app-mesh-inject#128 please be included here?

kishorj:
> Can the changes from [aws/aws-app-mesh-inject#128](https://github.com/aws/aws-app-mesh-inject/pull/128) please be included here?

Sure. I will include the changes in the next update.

pkg/appmeshinject/inject.go
```diff
@@ -0,0 +1,216 @@
...
+		jaegerInit, err := renderJaegerInitContainer(config.JaegerAddress, config.JaegerPort)
+		if err != nil {
+			return err
+		}
+		err = json.Unmarshal([]byte(jaegerInit), &container)
+		if err != nil {
+			return err
+		}
+		pod.Spec.InitContainers = append(pod.Spec.InitContainers, container)
+	}
```
mogren, line 213: These if-statements look almost the same. Could they be extracted to an interface, then implemented in each sidecar file instead? Something like:
```go
type SidecarRenderer interface {
   renderInitContainer() (corev1.Container, error)
}

func (sc *SidecarConfig) renderInitContainer() corev1.Container, error {}
```
(Se comment below for more details)

pkg/appmeshinject/sidecar.go
```diff
@@ -0,0 +1,158 @@
...
+	EnableDatadogTracing bool
+	DatadogAddress       string
+	DatadogPort          string
+	InjectXraySidecar    bool
+	EnableStatsTags      bool
+	EnableStatsD         bool
+}
+
+func renderSidecars(meta SidecarMeta) ([]string, error) {
+	var sidecars []string
```
mogren, line 121: I have a feeling that this package is kind of randomly divided. Like why is Envoy called Sidecar, and why is Xray special and inside this `renderSidecars` function, but Jaeger and DataDog and others are in separate files? Also, the logic to configure the metadata is very spread out. For example, `JaegerAddress` is used in four different files. 

Can't we have one config and one sidecar per file, and tie them together in `config.go`? Can `SidecarMeta` have the general config as separate fields, then a slice of a `SidecarConfig` type or similar where we add all configured sidecars? Then we can just loop over those configs in `InjectSideCars()`.

I understand that there are some dependencies between them, `EnableJaegerTracing` is used in the Envoy template for example, but maybe that is something we can make more clear as well, which ones depend on others?

kishorj: Thanks for your feedback.

Envoy is always injected unless sidecar injection is disabled entirely.

Jaegar and Datadog only have the initContainers and their purpose is to inject configuration to the envoy sidecar. There are three initContainers - 
1) proxyinit (or annotations if AppMesh CNI is enabled)
2) jaeger (if configured)
3) datadog (if configured)
and Two sidecars
1) Envoy
2) X-ray (if enabled)
I ported the changes mostly as-is from the inject repository.

Based on your previous comment as well, here is what I can do - 

type SidecarRenderer interface {
   renderInitContainer() (corev1.Container, error)
   renderContainer() (corev1.Container, error)
   renderConfigVolume() (corev1.Volume, error)
}
Each sidecar file will implement the interface, will return nil objects if not configured of not applicable. Then have the iterator in InjectSideCars() add the valid objects returned by the modules.

pkg/appmeshinject/inject.go
```diff
@@ -0,0 +1,197 @@
...
+	// Users should set fsGroup on the pod spec directly if a specific fsGroup is desired.
+	defaultFSGroup int64 = 1337
+)
+
+var (
+	config    Config
+	injectLog = ctrl.Log.WithName("appmesh_inject")
+)
+
+type SidecarRenderer interface {
```
M00nF1sh, line 28: i think it's might be better for this interface to just be like a mutatingwebhook.

like `mutate(pod *corev1.Pod) error` or `mutate(pod *corev1.pod) (*corev1.pod, error)`.
it's more flexiable.

kishorj: ok

M00nF1sh: something like this
```
func NewVNEnvoyMutator(config, vn) ...
type VNEnvoyMutator struct {
    config EnvoyMutatorConfig
    vn appmesh.virtualNode
}

func (m *EnvoyMutator) Mutate(pod *corev1.pod) error {
    pod.annotation[virtualnode] = vn.name
}


func injectPod(pod, vn) error {
    mutators = []Mutator{
        NewVNEnvoyMutator(config, vn, ...)
        ...
    }
    for _, mutator := range mutators{
        mutator.mutate(pod)
    }
}
```

pkg/appmeshinject/sidecar_utils.go
```diff
@@ -0,0 +1,136 @@
...
+		return v == "enabled"
+	}
+	//Fargate platform has appmesh-cni enabled by default
+	if v, ok := pod.GetLabels()[FargateProfileLabel]; ok {
+		return len(v) > 0
+	}
+	return false
+}
+
+func updateConfigFromPodAnnotations(config Config, pod *corev1.Pod) Config {
```
mogren, line 123: Nit: No unit tests for any of these functions

main.go
```diff
@@ -28,6 +28,7 @@ import (
 
 	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/mesh"
 	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/virtualnode"
+	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/appmeshinject"
```
M00nF1sh, line 31: nit, i'd suggest rename it to just be "inject" instead of "appmeshinject". but this works :D

pkg/appmeshinject/config.go
```diff
@@ -0,0 +1,107 @@
+package appmeshinject
+
+import (
+	"errors"
+	"flag"
+)
+
+type Config struct {
+	// Injetion Settings
```
M00nF1sh, line 9: injection

pkg/appmeshinject/config.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// If enabled, an fsGroup: 1337 will be injected in the absence of it within pod securityContext
+	// see https://github.com/aws/amazon-eks-pod-identity-webhook/issues/8 for more details
+	EnableIAMForServiceAccounts bool
+
+	// Sidecar settings
+	SidecarImage  string
+	SidecarCpu    string
+	SidecarMemory string
+	//MeshName      string
```
M00nF1sh, line 20: MeshName is not used, we can remove it

pkg/appmeshinject/config.go
```diff
@@ -0,0 +1,107 @@
...
+	EnableDatadogTracing        bool
+	DatadogAddress              string
+	DatadogPort                 string
+}
+
+// MultipleTracer checks if more than one tracer is configured.
+func MultipleTracer(config *Config) bool {
+	j := config.EnableJaegerTracing
+	d := config.EnableDatadogTracing
+	x := config.InjectXraySidecar
```
M00nF1sh, line 48: nit: since we rewrite this, it's a chance for us to normalize these flags. e.g. use InjectXXX or EnableXXX consistently.
I think we can also do it after merge this.

pkg/appmeshinject/jaeger.go
```diff
@@ -0,0 +1,76 @@
...
+             address: {{ .JaegerAddress }}
+             port_value: {{ .JaegerPort }}
+`
+
+const injectJaegerTemplate = `
+{
+  "command": [
+    "sh",
+    "-c",
+    "cat <<EOF >> /tmp/envoy/envoyconf.yaml{{ .Config }}EOF\n\ncat /tmp/envoy/envoyconf.yaml\n"
```
M00nF1sh, line 38: did this jaeger really depends on full content of this Config?
what will //tmp/envoy/envoyconf.yaml do

pkg/appmeshinject/jaeger.go
```diff
@@ -0,0 +1,76 @@
...
+`
+
+type JaegerMutator struct {
+}
+
+func (j *JaegerMutator) mutate(pod *corev1.Pod) error {
+	if !config.EnableJaegerTracing {
+		return nil
+	}
+	init, err := renderInitContainer("jaeger", jaegerTemplate, injectJaegerTemplate, config)
```
M00nF1sh, line 69: I'd favor use map[string]string instead of config for following reasons:
1. config contains too many info that may not need by this jaeger container.
2. it's not compile-time safe. (e.g. some one can rename some fields in config, but this will be broken undetected)

pkg/appmeshinject/jaeger.go
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+}
+`
+
+type JaegerMutator struct {
+}
+
+func (j *JaegerMutator) mutate(pod *corev1.Pod) error {
+	if !config.EnableJaegerTracing {
```
M00nF1sh, line 66: it's a antipattern to use global vars..
i'd favor component specific config been passed to mutator

pkg/appmeshinject/inject.go
```diff
@@ -0,0 +1,90 @@
...
+
+type PodMutator interface {
+	mutate(*corev1.Pod) error
+}
+
+func New(cfg Config) {
+	config = cfg
+}
+
+type AppMeshCNIMutator struct{}
```
M00nF1sh, line 33: seems cni mutator don't belong here..(it's should be it's own file)

pkg/appmeshinject/inject.go
```diff
@@ -0,0 +1,90 @@
...
+var (
+	config    Config
+	injectLog = ctrl.Log.WithName("appmesh_inject")
+)
+
+type PodMutator interface {
+	mutate(*corev1.Pod) error
+}
+
+func New(cfg Config) {
```
M00nF1sh, line 29: this global var usage seems weird=.=

M00nF1sh:
Thanks a lot @kishorj awesome work.
TBH, this is not my ideal quality, but i'm ok to merge it first to produce a testable binary, and we can improve this before GA via PRs.

pkg/appmeshinject/config.go
kishorj: ok

pkg/appmeshinject/jaeger.go
kishorj: This is the escaped Jaeger template that gets injected to the envoy config file - the term Config caused the confusion, I will rename it in the next commit.

pkg/appmeshinject/jaeger.go
kishorj: ok. I will investigate further.

pkg/appmeshinject/inject.go
kishorj: agree. I will refactor it.

pkg/appmeshinject/inject.go
kishorj: ok

pkg/appmeshinject/config.go
kishorj: will follow up

PR #167, opened 2020-04-20
initialize aws package

M00nF1sh:
# changes done
1. "add aws sdk throttle support" same change as https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/163. (already approve for master branch)
2. "add aws sdk metrics support"
     1. supports collect below metrics:
        * `aws_api_calls_total` (Total number of SDK API calls from the customer's code to AWS services)
        * `aws_api_call_duration_seconds` (Perceived latency from when your code makes an SDK call, includes retries)
        * `aws_api_call_retries` (Number of times the SDK retried requests to AWS services for SDK API calls)
        * `aws_api_requests_total` (Total number of HTTP requests that the SDK made)
        * `aws_api_request_duration_seconds` (Latency of an individual HTTP request to the service endpoint)
     2. difference than current metrics implementation used in master branch:
        1. more accurate, it now reports both latency/count from SDKClient's perspective and awsAPI's perspective. (a single SDK client apis call may result in multiple aws API calls)
       2. there isn't a separate `aws_api_errors` metric, error code are now models as a label per prometheus best practice.
       3. rely on session handler, so less intrusive into SDK wrapper.
3. "add thin sdk wrapper":
      1. the aws.cloud is meant as thin wrappers around aws SDK and aws related settings.
      2. the wrapper interface around sdk's api-iface is to allow us add helper method like `DescribeMeshesAsList` to handle pagination.
      3. the aws.cloud  have method to return interface to different services instead of embed them(there could be method conflict among services if embed them)

### DIFF
--- a/pkg/aws/cloud.go
+++ b/pkg/aws/cloud.go
@@ -1,0 +1,69 @@
+package aws
+
+import (
+	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/aws/metrics"
+	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/aws/services"
+	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/aws/throttle"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type Cloud interface {
+	// AppMesh provides API to AWS AppMesh
+	AppMesh() services.AppMesh
+	// CloudMap provides API to AWS CloudMap
+	CloudMap() services.CloudMap
+}
+
+// NewCloud constructs new Cloud implementation.
+func NewCloud(cfg CloudConfig, metricsRegisterer prometheus.Registerer) (Cloud, error) {
+	sess := session.Must(session.NewSession(aws.NewConfig()))
+	if cfg.ThrottleConfig != nil {
+		throttler := throttle.NewThrottler(cfg.ThrottleConfig)
+		throttler.InjectHandlers(&sess.Handlers)
+	}
+	if metricsRegisterer != nil {
+		metricsCollector, err := metrics.NewCollector(metricsRegisterer)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to initialize sdk metrics collector")
+		}
+		metricsCollector.InjectHandlers(&sess.Handlers)
+	}
+
+	if len(cfg.Region) == 0 {
+		metadata := services.NewEC2Metadata(sess)
+		region, err := metadata.Region()
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to introspect region from EC2Metadata, specify --aws-region instead if EC2Metadata is unavailable")
+		}
+		cfg.Region = region
+	}
+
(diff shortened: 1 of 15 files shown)


pkg/aws/cloud.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/aws/services"
+	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/aws/throttle"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type Cloud interface {
```
efe-selcuk, line 14: The name `Cloud` is my least favorite part of v1. It's basically a wrapper/facade around the SDK right?

efe-selcuk:
Can we break this PR down into multiple focused ones? I'm overwhelmed (even though part of it is copy-pasta.

pkg/aws/cloud.go
M00nF1sh: yep, do you have a better name?

M00nF1sh:
> Can we break this PR down into multiple focused ones? I'm overwhelmed (even though part of it is copy-pasta.

nice idea, updated :D

pkg/aws/cloud.go
mogren: AppMesh and Cloud map are both AWS specific, so how about just `AWSService`?

pkg/aws/cloud_config.go
```diff
@@ -0,0 +1,23 @@
...
+type CloudConfig struct {
+	// AWS Region for the kubernetes cluster
+	Region string
+	// Throttle settings for aws APIs
+	ThrottleConfig *throttle.ServiceOperationsThrottleConfig
+}
+
+func (cfg *CloudConfig) BindFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&cfg.Region, flagAWSRegion, "", "AWS Region for the kubernetes cluster")
+	fs.Var(cfg.ThrottleConfig, flagAWSAPIThrottle, "throttle settings for aws APIs, format: serviceID1:operationRegex1=rate:burst,serviceID2:operationRegex2=rate:burst")
```
mogren, line 22: Nit, `aws` -> `AWS`

M00nF1sh: the cloud is intended to hold all AWS specific configurations. e.g. both aws services's wrapper and eks specific configuration(if any).

I still prefer cloud, given it's widely used among projects, e.g.: 
1. https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L500
2. https://github.com/kubernetes/kops/blob/a9574284465216a26929d8ef5e5fa5ee8e69e92f/upup/pkg/fi/cloudup/awsup/aws_cloud.go
:D

M00nF1sh: updated aws -> AWS :D

dudicoco:
@M00nF1sh can you please explain the difference between `aws_api_calls_total` and `aws_api_requests_total`?

M00nF1sh:
@dudicoco 
the aws_api_calls_total is the number of calls from code. the aws_api_requests_total is the number of HTTP requests from SDK.

e.g. the application code(this controller) makes a single EC2:DescribeSubnet API call by invoke AWSSDKGo, the underlying AWSSDKGo may automatically retry multiple times(each time is a HTTP request) due to throttle/intermediate error.

PR #170, opened 2020-04-22
Dockerfile modification - yum update

achevuru:
*Issue #, if available:*

*Description of changes:*
Yum update as part of docker build..

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,7 @@ RUN make
 
 FROM amazonlinux:2
 RUN yum install -y ca-certificates
+RUN yum update -y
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -15,6 +15,7 @@ RUN make
 
 FROM amazonlinux:2
 RUN yum install -y ca-certificates
+RUN yum update -y
```
M00nF1sh, line 18: better to be below to save image size :D 
```
RUN yum update -y && \
    yum install -y ca-certificates && \
    yum clean all && \
    rm -rf /var/cache/yum
```

achevuru: Updated it..

PR #174, opened 2020-04-23
V1beta2 gateway

rajal-amzn:
*Issue #, if available:*
[aws-app-mesh-roadmap #37](https://github.com/aws/aws-app-mesh-roadmap/issues/37)

*Description of changes:*
Adding mutator and validator for Virtual Gateway API. The mutator adds the immutable fields (awsName and meshId) whereas the validator validates that the immutable fields are not changed between updates.

Note for reviewer:
The PR is separated into 2 commits, only the 1st one needs to be reviewed while the second one is auto-generated.

### DIFF
--- a/webhooks/appmesh/virtualgateway_mutator_test.go
+++ b/webhooks/appmesh/virtualgateway_mutator_test.go
@@ -1,0 +1,216 @@
+package appmesh
+
+import (
+	"context"
+	appmesh "github.com/aws/aws-app-mesh-controller-for-k8s/apis/appmesh/v1beta2"
+	mock_mesh "github.com/aws/aws-app-mesh-controller-for-k8s/mocks/aws-app-mesh-controller-for-k8s/pkg/mesh"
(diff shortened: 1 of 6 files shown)


webhooks/appmesh/virtualgateway_mutator_test.go
```diff
@@ -0,0 +1,216 @@
...
+		want    *appmesh.VirtualGateway
+		wantErr error
+	}{
+		{
+			name: "VirtualGateway didn't specify awsName",
+			args: args{
+				vGateway: &appmesh.VirtualGateway{
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace: "my-ns",
+						Name:      "my-vg",
```
karanvasnani, line 31: I think this should also include `UID`.

M00nF1sh: it's ok to don't have UID here per this test case's purpose :D

PR #176, opened 2020-04-27
virtualnode controller

M00nF1sh:
## This PR adds virtualNode controller and all  infrastructure functionalities supports it.

Note: It should be served as high-level style guideline and reference implementation for all other controllers

### Object lifecycle
1. customer creates a mesh, which selects namespaces.
2. customer creates namespaced objects in namespace.
    1. mutating webhook will find corresponding mesh for object's namespace and persist it as spec.meshRef on objects.
3. meshController detects mesh creation:
    1. two finalizer will be added to mesh:
        * `finalizers.appmesh.k8s.aws/mesh-members`: responsible to block mesh deletion unless all namespaced appmesh CRs referencing this mesh have been deleted. 
             Note: The controller **should not automatically delete** namespaced appmesh CRs upon mesh deletion due to following reasons:
             *  to match AppMesh API's behavior, where mesh deletion requires to clean up objects in mesh first.
             * to avoid race if customer have controllers to manage AppMesh objects. (e.g. flagger)
        * `finalizers.appmesh.k8s.aws/aws-resources`: responsible to block mesh deletion unless the mesh in AppMesh API is deleted.
     2. an mesh will be created through AppMesh API 
4. (happens simultaneously with #3) virtualNodeController detects virtualNodes creation:
    1. one finalizer will be added :
        * `finalizers.appmesh.k8s.aws/aws-resources`: responsible to block mesh deletion unless the virtualNode in AppMesh API is deleted.
    2. validates in-cluster dependencies for virtualNode:
       * mesh must be in 'active" state: which means meshController have finished create mesh in AppMesh API
       * virtual-services must exists and belong to same mesh
    3. if in-cluster dependencies is not fulfilled, re-queue the virtualNode.
    4. as an optimization for in-cluster dependencies handling, when mesh `turn active` or virtualService creation, enqueue related virtualNode immediately. 
        Note: we can possibly avoid "re-queue the virtualNode" in step3 above if we implemented all optimization for in-cluster dependencies handling. however, since validates in-cluster dependencies is cheap and always happens first, we don't need to.
    5. create virtualnode.
5. (happens simultaneously with #3), virtualServiceController/virtualRouterController detects virtualService/virtualRouter creation:
    1. one finalizer will be added :
        * `finalizers.appmesh.k8s.aws/aws-resources`: responsible to block mesh deletion unless the virtualService/virtualRouter in AppMesh API is deleted.
     2. validates in-cluster dependencies for virtualService/virtualRouter:
        * mesh must be in 'active" state: which means meshController have finished create mesh in AppMesh API
       * virtualNode must exists and active. (which means virtualNodeController have finished create virtualNode in AppMesh API)
     3. if in-cluster dependencies is not fulfilled, re-queue the virtualService/virtualRouter.
     4. as an optimization for in-cluster dependencies handling, when mesh `turn active` or virtualNode "turn active", enqueue related virtualService/virtualRouter immediately.
     5. virtualService/virtualRouter.
6. customer updates namespaced objects in namespace.
    similar follow for object creation applies.
7. customer deletes namespaced objects in namespace.
    reverse order of creation flow applies. dependencies should be respected, e.g. virtualNode cannot be deleted if there is any virtualService referencing it.
8. customer deletes mesh.

### The commits is break down into 7 independent commits:
1. `add finalizer utilities`: add a utility to deal with finalizers.
     Two finalizer is planed:
     * `finalizers.appmesh.k8s.aws/aws-resources`: responsible for block CR deletion to cleanup AWS resources created.
    *  `finalizers.appmesh.k8s.aws/mesh-members`: responsible for block mesh deletion if mesh have members. mesh deletion should' not delete the member CR
2. `add equality compare options`: add utilities to handle compare two objects:
     * `IgnoreLeftHandUnset ` is used to ignore unset values in left-hand side arg when doing compare. e.g. when compare(desiredAWSSpec, actualAWSSpec), it allows AppMesh to fill default field in actualAWSSpec if the field is nil in desiredAWSSpec.
     * `IgnoreFakeClientPopulatedFields` is used for test cases to compare two apiObjects while ignore TypeMeta and ResourceVersion, as these two will be populated by fakeClient.
3. `add mesh & virtualservice reference handling components`: utility components to handle references.
4. `add runtime error handling components`: components to handle runtime errors. It allows controller-code to just return specific error type to control re-queue behavior.
5. `add type versions from crd to sdk`: add type conversion functionalities. Note: It's explicitly using the `Convert_CRD_<Type>_To_SDK_<Type>` function signature(instead of `ConvertCRDTypeToSDKType`) to be clear. we could auto-generate these in the future
6. `add virtualNode resource manager`: which is responsible for dealing virtualNode creation/update/delete in AppMesh API.
7. `add virtualNode controller & mesh eventHandler`: controller and event handler to hook all together.


pkg/conversions/virtualnode_types_conversion.go
```diff
@@ -0,0 +1,329 @@
+package conversions
```
mogren, line 1: The function names in this file is not very go like, it's recommended to use [MixedCaps](https://golang.org/doc/effective_go.html#mixed-caps).

Also, do all these functions have to be public?

pkg/equality/ignore_unset.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"github.com/google/go-cmp/cmp"
+	"reflect"
+)
+
+// IgnoreLeftHandUnset is an option that ignores fields that are unset on the
+// left hand side of a comparison.
+// Note:
+//	1. for map and slices, only nil value is considered to be unset, non-nil but empty is not considered as unset.
+//  2. for struct pointers, nil value is considered to be unset
```
mogren, line 12: Nit: Indentation

pkg/virtualnode/resource_manager.go
```diff
@@ -0,0 +1,276 @@
...
+		Tags:            nil,
+		VirtualNodeName: vn.Spec.AWSName,
+	})
+	if err != nil {
+		return nil, err
+	}
+	return resp.VirtualNode, nil
+}
+
+func (m *defaultResourceManager) updateSDKVirtualNode(ctx context.Context, sdkVN *appmeshsdk.VirtualNodeData, ms *appmesh.Mesh, vn *appmesh.VirtualNode, vsByRef map[appmesh.VirtualServiceReference]*appmesh.VirtualService) (*appmeshsdk.VirtualNodeData, error) {
```
mogren, line 190: This is only called in the `Reconcile`, but it might be worth a comment that `IgnoreLeftHandUnset` is used when checking for equality. If the `desiredSDKVNSpec` is missing some settings, what does that mean?

pkg/virtualservice/utils.go
```diff
@@ -0,0 +1,17 @@
+package virtualservice
+
+import (
+	appmesh "github.com/aws/aws-app-mesh-controller-for-k8s/apis/appmesh/v1beta2"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// IsVirtualServiceActive checks whether given virtualService is active.
+// virtualService is active when its VirtualServiceActive condition equals true.
+func IsVirtualServiceActive(vs *appmesh.VirtualService) bool {
```
mogren, line 10: Is this used anywhere except in the test?

pkg/virtualnode/resource_manager_test.go
```diff
@@ -0,0 +1 @@
+package virtualnode
```
mogren, line 1: No tests? 🤨

M00nF1sh: yeah, plan to add after initial review, adding :D

pkg/virtualservice/utils.go
M00nF1sh: no, it's not used anywhere. (but will be used in future PRs) :D

pkg/virtualnode/resource_manager.go
M00nF1sh: if an optional field is not set, AWS will provide default settings(and will be in actualSDKVNSpec). And compare will fail always.
This allows to equality to still be true if specific optional settings is omitted, which allows AWS sever-side to provide default settings.

pkg/equality/ignore_unset.go
M00nF1sh: :D fixing

pkg/conversions/virtualnode_types_conversion.go
M00nF1sh: They don't need to be all public, but I want to follow same naming pattern. (and Public allows to convert specific sub structures if it's needed).
I'm use this naming pattern to make conversion src and dest explicitly, and also follow k8s's pattern.
Quoted from https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#edit-version-conversions
> Since auto-generated conversion functions are using manually written ones, those manually written should be named with a defined convention, i.e. a function converting type X in pkg a to type Y in pkg b, should be named: convert_a_X_To_b_Y.

pkg/runtime/errors.go
```diff
@@ -0,0 +1,68 @@
...
+// instruct controller-runtime to requeue the processing item without been logged as error.
+func NewRequeueError(err error) *RequeueError {
+	return &RequeueError{
+		err: err,
+	}
+}
+
+// NewRequeueAfterError constructs new RequeueAfterError to
+// instruct controller-runtime to requeue the processing item after specified duration without been logged as error.
+func NewRequeueAfterError(err error, duration time.Duration) *RequeueAfterError {
```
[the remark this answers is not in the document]
M00nF1sh: It's not used so far, but added as complementary to RequeueError.
So functionality related to controller runtime's ctrl.Result can be fully supported.

pkg/mesh/reference_resolver.go
```diff
@@ -0,0 +1,45 @@
...
+
+var _ ReferenceResolver = &defaultReferenceResolver{}
+
+// defaultReferenceResolver implements ReferenceResolver
+type defaultReferenceResolver struct {
+	k8sClient client.Client
+	log       logr.Logger
+}
+
+// Resolve returns an mesh CR based on meshRef
```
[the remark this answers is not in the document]
M00nF1sh: will change

pkg/conversions/virtualnode_types_conversion_test.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	appmesh "github.com/aws/aws-app-mesh-controller-for-k8s/apis/appmesh/v1beta2"
+	"github.com/aws/aws-sdk-go/aws"
+	appmeshsdk "github.com/aws/aws-sdk-go/service/appmesh"
+	"github.com/stretchr/testify/assert"
+	"k8s.io/apimachinery/pkg/conversion"
+	"testing"
+)
+
+func TestConvert_CRD_TLSValidationContextACMTrust_To_SDK_TLSValidationContextACMTrust(t *testing.T) {
```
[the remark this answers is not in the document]
M00nF1sh: > That's quite a lot of changes in one PR. The only major concern I have is the function names in the `conversions` package.

It's separated into multiple independent commits. So they can be reviewed separately while still gives reviewer the ability to peak the high-level architecture and usage.
(btw, even in this PR, you asked how a unused function will be used, submit this in one PR instead of 7 PR is to reduce such questions, 🤣 )

pkg/conversions/virtualnode_types_conversion.go
```diff
@@ -0,0 +1,329 @@
...
+		if err := Convert_CRD_AccessLog_To_SDK_AccessLog(crdObj.AccessLog, sdkObj.AccessLog, scope); err != nil {
+			return err
+		}
+	} else {
+		sdkObj.AccessLog = nil
+	}
+	return nil
+}
+
+func Convert_CRD_VirtualNodeSpec_To_SDK_VirtualNodeSpec(crdObj *appmesh.VirtualNodeSpec, sdkObj *appmeshsdk.VirtualNodeSpec, scope conversion.Scope) error {
```
fawadkhaliq, line 275: It'd be good to add testing (with 0, 1, >1 listeners, backends) for `Convert_CRD_VirtualNodeSpec_To_SDK_VirtualNodeSpec` as well

pkg/mesh/reference_resolver_test.go
```diff
@@ -0,0 +1,110 @@
...
+		{
+			name: "mesh can be resolved when name and UID matches",
+			env: env{
+				meshes: []*appmesh.Mesh{meshUIDMatches},
+			},
+			args: args{
+				meshRef: appmesh.MeshReference{
+					Name: "my-mesh",
+					UID:  "a385048d-aba8-4235-9a11-4173764c8ab7",
+				},
```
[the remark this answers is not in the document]
M00nF1sh: they are always set, and their type is "string", so it's safe already :D

pkg/virtualnode/enqueue_requests_for_mesh_events.go
```diff
@@ -0,0 +1,71 @@
...
+func (h *enqueueRequestsForMeshEvents) Generic(e event.GenericEvent, queue workqueue.RateLimitingInterface) {
+	// no-op
+}
+
+func (h *enqueueRequestsForMeshEvents) enqueueVirtualNodesForMesh(ctx context.Context, queue workqueue.RateLimitingInterface, ms *appmesh.Mesh) {
+	vnList := &appmesh.VirtualNodeList{}
+	if err := h.k8sClient.List(ctx, vnList); err != nil {
+		h.log.Error(err, "failed to enqueue virtualNodes for mesh events",
+			"mesh", k8s.NamespacedName(ms))
+		return
```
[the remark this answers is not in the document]
M00nF1sh: yes, and this is used as an optimization. The controller replies on informer & APIServer to work correctly.
Also this list is based on cache.

pkg/virtualnode/resource_manager.go
```diff
@@ -0,0 +1,276 @@
...
+func (m *defaultResourceManager) findSDKVirtualNode(ctx context.Context, ms *appmesh.Mesh, vn *appmesh.VirtualNode) (*appmeshsdk.VirtualNodeData, error) {
+	resp, err := m.appMeshSDK.DescribeVirtualNodeWithContext(ctx, &appmeshsdk.DescribeVirtualNodeInput{
+		MeshName:        ms.Spec.AWSName,
+		MeshOwner:       ms.Spec.MeshOwner,
+		VirtualNodeName: vn.Spec.AWSName,
+	})
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "NotFoundException" {
+			return nil, nil
+		}
```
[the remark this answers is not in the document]
M00nF1sh: provide a uniformed interface to caller. e.g. use nil(resource), nil(error) to represent "not found", otherwise, caller needs to check for the error.

pkg/conversions/virtualnode_types_conversion.go
M00nF1sh: sure

controllers/appmesh/virtualnode_controller.go
```diff
@@ -18,36 +18,104 @@ package controllers
...
+	return nil
+}
+
+func (r *virtualNodeReconciler) addFinalizers(ctx context.Context, vn *appmesh.VirtualNode) error {
+	if k8s.HasFinalizer(vn, k8s.FinalizerAWSResources) {
+		return nil
+	}
+	oldVN := vn.DeepCopy()
+	controllerutil.AddFinalizer(vn, k8s.FinalizerAWSResources)
+	if err := r.k8sClient.Patch(ctx, vn, client.MergeFrom(oldVN)); err != nil {
```
mogren, line 105: Why is the `client.MergeFrom()` needed both here and when deleting finalizers?

pkg/virtualnode/resource_manager.go
mogren: Thanks, that explains it! Might be worth adding as a go doc comment since this is not obvious:
```go
// updateSDKVirtualNode updates the SDK Virtual Node. If an optional field is not set, AWS will provide default 
// settings that will be in actualSDKVNSpec. We use IgnoreLeftHandUnset when doing the equality check here to
// allow AWS sever-side to provide default settings.

```

pkg/mesh/reference_resolver.go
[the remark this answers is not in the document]
mogren: Will you really? 😄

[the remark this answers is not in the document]
M00nF1sh: i think i changed in latest commit :D

controllers/appmesh/virtualnode_controller.go
mogren: Ok, discussed offline. It's an optimization to generate and apply the diff between the oldVN and the change that will be applied by adding or deleting the Finalizer, not reapply the whole config.

pkg/virtualnode/resource_manager.go
M00nF1sh: thanks, added comments :D

PR #191, opened 2020-05-04
VirtualNode Controller - CloudMap

achevuru:
*Issue #, if available:*

*Description of changes:*
VirtualNode Controller - CloudMap


controllers/appmesh/cloudmap_controller.go
```diff
@@ -0,0 +1,888 @@
...
+	if !vNode.DeletionTimestamp.IsZero() {
+		return r.deleteCloudMapResources(ctx, vNode)
+	}
+	return r.reconcileVirtualNodeWithCloudMap(ctx, vNode)
+}
+
+func (r *cloudMapReconciler) reconcileVirtualNodeWithCloudMap(ctx context.Context, vNode *appmesh.VirtualNode) error {
+
+	if !r.isCloudMapEnabledForVirtualNode(vNode) {
+		return errors.Errorf("CloudMap not enabled for virtualNode: %s", vNode.Name)
```
M00nF1sh, line 152: this shouldn't be error. when return error, it will backoff reconcile.
these are valid configurations, we should just return nil.

controllers/appmesh/cloudmap_controller.go
```diff
@@ -0,0 +1,888 @@
...
+	}
+
+	if !r.isPodSelectorDefinedForVirtualNode(vNode) {
+		return errors.Errorf("PodSelector not defined for virtualNode: %s", vNode.Name)
```
M00nF1sh, line 156: same as above

controllers/appmesh/cloudmap_controller.go
```diff
@@ -0,0 +1,888 @@
...
+	}
+
+	cloudmapConfig := vNode.Spec.ServiceDiscovery.AWSCloudMap
```
M00nF1sh, line 159: nit, maybe `cloudMapConfig`

controllers/appmesh/cloudmap_controller.go
```diff
@@ -0,0 +1,888 @@
...
+	}
+	r.serviceIDCache.Add(serviceItem)
+	return &serviceItem.value, nil
+}
+
+func (r *cloudMapReconciler) createCloudMapService(ctx context.Context,
+	cloudmapConfig *appmesh.AWSCloudMapServiceDiscovery, creatorRequestID string) (*cloudMapServiceSummary, error) {
+
+	key := r.serviceCacheKey(cloudmapConfig)
+	if serviceSummary, _ := r.getCloudMapServiceFromCache(ctx, key); serviceSummary != nil {
```
M00nF1sh, line 605: why is error ignored, when will it be error?

controllers/appmesh/cloudmap_controller.go
```diff
@@ -0,0 +1,888 @@
...
+		return err
+	}
+
+	r.log.V(2).Info("Pods found matching this VNode: ", "Total Pods: ", len(podsList.Items))
+	if err := r.finalizerManager.AddFinalizers(ctx, vNode, k8s.FinalizerAWSCloudMapResources); err != nil {
+		r.log.Error(err, "..while adding cloudmap resource finalizer")
+		return err
+	}
+
+	_, err := r.createCloudMapService(ctx, cloudmapConfig, string(creatorRequestID))
```
M00nF1sh, line 178: better to be named as "reconcileCloudMapService" or "getOrCreateCloudMapService)

controllers/appmesh/cloudmap_controller.go
```diff
@@ -0,0 +1,888 @@
...
+		}
+
+		instanceId := r.podToInstanceID(&pod)
+		if instanceId == "" {
+			r.log.Info("No IP Address assigned to Pod:", pod.Name, "..Skipping for now")
+			continue
+		}
+
+		instanceInfo := &cloudMapInstanceInfo{}
+		if r.isPodReady(&pod) {
```
M00nF1sh, line 198: better to separate a component to get backend pods for a virtualNodes.
some thing like ```
type virtualNodeEndpointResolver interface {
     Resolve(ctxcontext.context, vn *appmesh.VirtualNode) (ready []*corev1.Pod, notReady[]corev1.Pod, ignoredPods []corev1.Pod)
}
```

File: controllers/appmesh/cloudmap_controller.go [hunk 7]
```diff
@@ -0,0 +1,888 @@
...
+		instanceInfo.virtualNodeName = vNode.Name
+		instanceInfo.meshName = vNode.Name
+		instanceInfo.labels = make(map[string]string)
+		for label, v := range pod.Labels {
+			instanceInfo.labels[label] = v
+		}
+		serviceInstanceSummary[instanceId] = instanceInfo
+	}
+
+	if err := r.updateCloudMapServiceInstances(ctx, cloudmapConfig, serviceInstanceSummary, string(creatorRequestID)); err != nil {
```
M00nF1sh, line 216: better to have a component dedicated to register/deregister instances from cloudmap

pkg/virtualnode/enqueue_requests_for_pod_events.go
```diff
@@ -0,0 +1,95 @@
+package virtualnode
```
M00nF1sh, line 1: this should be in /pkg/cloudmap since it's used by the cloudmap controller.

pkg/virtualnode/enqueue_requests_for_pod_events.go
```diff
@@ -0,0 +1,95 @@
...
+	// cloudmap reconcile depends Virtualnode Pod Selector labels and if there is an update to a Pod
+	// we need to check if there is any change w.r.t VirtualNode it belongs.
+	oldPod := e.ObjectOld.(*corev1.Pod)
+	newPod := e.ObjectNew.(*corev1.Pod)
+
+	oldPodIsReady := false
+	newPodIsReady := false
+
+	conditions := (&oldPod.Status).Conditions
+	for i := range conditions {
```
M00nF1sh, line 48: better to have utilities to test whether pod is ready, and reuse it here.

controllers/appmesh/cloudmap_controller.go
```diff
@@ -0,0 +1,888 @@
...
+	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/k8s"
+)
+
+// CloudMapReconciler reconciles a VirtualNode pod instance to CloudMap Service
+type cloudMapReconciler struct {
+	k8sClient                   client.Client
+	log                         logr.Logger
+	cloudMapSDK                 awsservices.CloudMap
+	finalizerManager            k8s.FinalizerManager
+	virtualNodeEndPointResolver cloudmap.VirtualNodeEndpointResolver
```
M00nF1sh, line 50: nit: endpoint :D

controllers/appmesh/cloudmap_controller.go
```diff
@@ -0,0 +1,888 @@
...
+		return err
+	}
+
+	_, err := r.getOrCreateCloudMapService(ctx, cloudMapConfig, string(creatorRequestID))
+	if err != nil {
+		r.log.Error(err, "failed to create cloudMap Service for", "vNode: ", vNode.Name)
+		return err
+	}
+
+	readyPods, notReadyPods, ignoredPods, perr := r.virtualNodeEndPointResolver.ResolveCloudMapEndPoints(ctx, vNode)
```
M00nF1sh, line 178: perr -> err?

controllers/appmesh/cloudmap_controller.go
```diff
@@ -0,0 +1,888 @@
...
+		r.log.Error(perr, "failed to get pods for", "vNode: ", vNode.Name)
+		return perr
+	}
+
+	r.log.V(1).Info("EndPoints: ", "Ready Pods Count: ", len(readyPods.Items),
+		"Unready Pods Count: ", len(notReadyPods.Items), "Ignored Pods Count: ", len(ignoredPods.Items))
+	if len(readyPods.Items) == 0 && len(notReadyPods.Items) == 0 {
+		if len(ignoredPods.Items) > 0 {
+			//Handle the scenario where the deployment behind the Virtualnode is deleted
+			if err := r.deleteAWSCloudMapServiceInstances(ctx, cloudMapConfig); err != nil {
```
M00nF1sh, line 189: we should just always update cloudmap if `readyPods+notReadypods` don't match our cached cloudmap status.
K8s can miss events, so it's possible to only get events when all pods are gone. (including ignorePods)

pkg/cloudmap/enqueue_requests_for_pod_events.go
```diff
@@ -0,0 +1,80 @@
...
+	pod *corev1.Pod) {
+	vnList := &appmesh.VirtualNodeList{}
+	if err := h.k8sClient.List(ctx, vnList); err != nil {
+		h.log.Error(err, "failed to enqueue virtualNodes for pod events",
+			"Pod", k8s.NamespacedName(pod))
+		return
+	}
+
+	for _, vn := range vnList.Items {
+		if vn.Spec.PodSelector == nil {
```
M00nF1sh, line 72: we don't need this check.
LabelSelectorAsSelector handles nil case correctly, we can just use it, and use UT to guarantee the behavior don't change.

pkg/cloudmap/enqueue_requests_for_pod_events.go
```diff
@@ -0,0 +1,80 @@
...
+			return
+		}
+		selector, _ := metav1.LabelSelectorAsSelector(vn.Spec.PodSelector)
```
M00nF1sh, line 75: we should check error here. and ignore the virtualnode if it's error.
we shouldn't use any value before error is checked. (unless it's guaranteed to be succeed, e.g. compile time regex)

controllers/appmesh/cloudmap_controller.go
achevuru: will change it

pkg/cloudmap/enqueue_requests_for_pod_events.go
achevuru: will remove the check

controllers/appmesh/cloudmap_controller.go
achevuru: Yeah, don't why i named it perr :)...

pkg/cloudmap/enqueue_requests_for_pod_events_test.go
```diff
@@ -0,0 +1,358 @@
...
+					"app": "testapp",
+				},
+			},
+		},
+	}
+
+	type env struct {
+		virtualNodes []*appmesh.VirtualNode
+	}
+	type args struct {
```
M00nF1sh, line 40: let's have a dedicated test case for enqueueVirtualNodesForPods, which have check on pod's selector == nil case.

achevuru: ok

pkg/cloudmap/enqueue_requests_for_pod_events.go
achevuru: will do

pkg/cloudmap/enqueue_requests_for_pod_events.go
```diff
@@ -0,0 +1,80 @@
...
+// Generic is called in response to an event of an unknown type or a synthetic event triggered as a cron or
+// external trigger request
+func (h *enqueueRequestsForPodEvents) Generic(e event.GenericEvent, queue workqueue.RateLimitingInterface) {
+	// no-op
+}
+
+func (h *enqueueRequestsForPodEvents) enqueueVirtualNodesForPods(ctx context.Context, queue workqueue.RateLimitingInterface,
+	pod *corev1.Pod) {
+	vnList := &appmesh.VirtualNodeList{}
+	if err := h.k8sClient.List(ctx, vnList); err != nil {
```
M00nF1sh, line 65: we should only list vn in the pod's namespace

pkg/cloudmap/enqueue_requests_for_pod_events.go
```diff
@@ -0,0 +1,80 @@
...
+	log       logr.Logger
+}
+
+// Create is called in response to an create event
+func (h *enqueueRequestsForPodEvents) Create(e event.CreateEvent, queue workqueue.RateLimitingInterface) {
+	h.enqueueVirtualNodesForPods(context.Background(), queue, e.Object.(*corev1.Pod))
+}
+
+// Update is called in response to an update event
+func (h *enqueueRequestsForPodEvents) Update(e event.UpdateEvent, queue workqueue.RateLimitingInterface) {
```
M00nF1sh, line 38: ideally we should mirror endpointController's behavior here. (can be in follow up PR).
https://github.com/kubernetes/kubernetes/blob/3491129d348eb6fb5e4274112df5a946ded740af/pkg/controller/util/endpoint/controller_utils.go#L187
Things missing here:
1. we should check whether [pod's deletionTimestamp differs](https://github.com/kubernetes/kubernetes/blob/3491129d348eb6fb5e4274112df5a946ded740af/pkg/controller/util/endpoint/controller_utils.go#L168) we used deletionTimestamp in endpointResolver
2. we should check whether pod's labels changed, if pod labels changed.  if pod ready also changed, enqueue the union of vn(oldPod), vn(newPod). if pod ready didn't change, enqueue vn(oldPod), vn(newPod) (if they differs)

pkg/cloudmap/virtualnode_endpoint_resolver.go
```diff
@@ -0,0 +1,73 @@
...
+	appmesh "github.com/aws/aws-app-mesh-controller-for-k8s/apis/appmesh/v1beta2"
+	awsservices "github.com/aws/aws-app-mesh-controller-for-k8s/pkg/aws/services"
+)
+
+type VirtualNodeEndpointResolver interface {
+	ResolveCloudMapEndPoints(ctx context.Context,
+		vn *appmesh.VirtualNode) (corev1.PodList, corev1.PodList, corev1.PodList, error)
+}
+
+type EndPointResolver struct {
```
M00nF1sh, line 19: nit, End**p**ointResolver

pkg/cloudmap/utils.go
```diff
@@ -0,0 +1,34 @@
+package cloudmap
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+func IsPodReady(pod *corev1.Pod) bool {
```
M00nF1sh, line 7: let's rename it as "IsPodContainersReady(pod *corev1.Pod) bool and check corev1.ContainersReady condition instead.
corev1.PodReady won't be true once we add podReadinessGate. (it will only be true when our podReadinessGate condition be true, however we need to add the pod into cloudmap for our readinessgate pass)

pkg/cloudmap/utils.go
```diff
@@ -0,0 +1,34 @@
...
+		return pod.Status.Phase != corev1.PodFailed && pod.Status.Phase != corev1.PodSucceeded
+	case corev1.RestartPolicyOnFailure:
+		return pod.Status.Phase != corev1.PodSucceeded
+	default:
+		return true
+	}
+}
+
+func PodToInstanceID(pod *corev1.Pod) string {
+	if pod.Status.PodIP == "" {
```
M00nF1sh, line 30: seems, seems it's just return pod.Status.PodIP..

pkg/cloudmap/virtualnode_endpoint_resolver.go
```diff
@@ -0,0 +1,73 @@
...
+
+	for _, pod := range podsList.Items {
+		if pod.DeletionTimestamp != nil {
+			e.log.Info("vNode:", "Pod is being deleted: ", pod.Name)
+			ignoredPods.Items = append(ignoredPods.Items, pod)
+			continue
+		}
+
+		instanceId := PodToInstanceID(&pod)
+		if instanceId == "" {
```
M00nF1sh, line 58: we should just check pod.ip.status here. This component should be independent from cloudMap. Its sole purpose is to find set of pods belong to virtualNode.  which maybe added to cloudmap or some other service

pkg/cloudmap/virtualnode_endpoint_resolver.go
```diff
@@ -0,0 +1,73 @@
...
+			ignoredPods.Items = append(ignoredPods.Items, pod)
+			e.log.Info("No IP Address assigned to Pod:", pod.Name, "..Skipping for now")
+			continue
+		}
+
+		if IsPodReady(&pod) {
+			readyPods.Items = append(readyPods.Items, pod)
+		} else if ShouldPodBeRegisteredWithCloudMapService(&pod) {
```
M00nF1sh, line 66: this should just be "shouldPodBeInEndpoints", and embed inside this component instead of in utils.
It have nothing to do with cloudMap.  just a general component to find endpoints for a virtualNode.

controllers/appmesh/cloudmap_controller.go
achevuru: renamed it

controllers/appmesh/cloudmap_controller.go
achevuru: done

controllers/appmesh/cloudmap_controller.go
achevuru: done

pkg/virtualnode/enqueue_requests_for_pod_events.go
achevuru: done

pkg/virtualnode/enqueue_requests_for_pod_events.go
achevuru: ok

controllers/appmesh/cloudmap_controller.go
achevuru: ok

pkg/cloudmap/enqueue_requests_for_pod_events.go
achevuru: done

pkg/cloudmap/utils.go
achevuru: Yes, just a wrapper to extract the instance id used by cloudmap. Right now, it is pod IP but if it does change this will act as an easy way to propagate the change.

pkg/cloudmap/virtualnode_endpoint_resolver.go
achevuru: done

pkg/cloudmap/virtualnode_endpoint_resolver.go
achevuru: changed it

pkg/cloudmap/virtualnode_endpoint_resolver.go
achevuru: ok

pkg/cloudmap/utils.go
achevuru: ok

controllers/appmesh/cloudmap_controller.go
```diff
@@ -0,0 +1,559 @@
...
+func (r *cloudMapReconciler) reconcile(req ctrl.Request) error {
+	ctx := context.Background()
+	log := r.log.WithValues("Virtualnode", req.NamespacedName)
+
+	vNode := &appmesh.VirtualNode{}
+	if err := r.k8sClient.Get(ctx, req.NamespacedName, vNode); err != nil {
+		return client.IgnoreNotFound(err)
+	}
+
+	log.V(2).Info("VNode: ", "awsName: ", vNode.Spec.AWSName,
```
M00nF1sh, line 132: lets rename this log statement.

controllers/appmesh/cloudmap_controller.go
```diff
@@ -0,0 +1,559 @@
...
+	cloudMapSDK                 awsservices.CloudMap
+	finalizerManager            k8s.FinalizerManager
+	virtualNodeEndpointResolver cloudmap.VirtualNodeEndpointResolver
+	cloudMapInstanceReconciler  cloudmap.CloudMapInstanceReconciler
+	nameSpaceIDCache            *cache.LRUExpireCache
+	serviceIDCache              *cache.LRUExpireCache
+	enqueueRequestsForPodEvents handler.EventHandler
+}
+
+type cloudmapNamespaceCacheItem struct {
```
M00nF1sh, line 57: i believe these two cache item is not used now :D

controllers/appmesh/cloudmap_controller.go
```diff
@@ -0,0 +1,559 @@
...
+			return err
+		}
+		if err := r.finalizerManager.RemoveFinalizers(ctx, vNode, k8s.FinalizerAWSCloudMapResources); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (r *cloudMapReconciler) getCloudMapNameSpaceFromCache(ctx context.Context,
```
M00nF1sh, line 196: nit: Name**s**pace

controllers/appmesh/cloudmap_controller.go
```diff
@@ -0,0 +1,559 @@
...
+		return nil, nil
+	}
+	return &namespaceSummary, err
+}
+
+func (r *cloudMapReconciler) getCloudMapNameSpace(ctx context.Context,
+	cloudMapConfig *appmesh.AWSCloudMapServiceDiscovery) (cloudMapNamespaceSummary, error) {
+	key := r.namespaceCacheKey(cloudMapConfig)
+
+	if namespaceSummary, _ := r.getCloudMapNameSpaceFromCache(ctx, key); namespaceSummary.NamespaceID != "" {
```
M00nF1sh, line 250: tbh, i'd favor to just getCloudMapNamespaceFromCache(ctx, key) to return namespaceSummary & exists.

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
-  newTag: latest
+  newName: 928111597794.dkr.ecr.us-east-1.amazonaws.com/amazon/app-mesh-controller
```
M00nF1sh, line 7: remove these

controllers/appmesh/cloudmap_controller.go
achevuru: ok

controllers/appmesh/cloudmap_controller.go
achevuru: can change it

config/manager/kustomization.yaml
achevuru: will remove

PR #194, opened 2020-05-04
mesh events handler for vr/vs

M00nF1sh:
Add mesh event handler for virtualRouter and virtualService.
These are merely copy-paste of mesh event handler for virtualNode :D 

### DIFF
--- a/pkg/virtualrouter/enqueue_requests_for_mesh_events.go
+++ b/pkg/virtualrouter/enqueue_requests_for_mesh_events.go
@@ -1,0 +1,71 @@
+package virtualrouter
+
+import (
+	"context"
+	appmesh "github.com/aws/aws-app-mesh-controller-for-k8s/apis/appmesh/v1beta2"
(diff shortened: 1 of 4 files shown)


pkg/virtualrouter/enqueue_requests_for_mesh_events.go
```diff
@@ -0,0 +1,71 @@
...
+
+var _ handler.EventHandler = (*enqueueRequestsForMeshEvents)(nil)
+
+type enqueueRequestsForMeshEvents struct {
+	k8sClient client.Client
+	log       logr.Logger
+}
+
+// Create is called in response to an create event
+func (h *enqueueRequestsForMeshEvents) Create(e event.CreateEvent, queue workqueue.RateLimitingInterface) {
```
mogren, line 31: So create and delete calls are no-ops because that is handled outside of this reconcile loop?

M00nF1sh: because virtualNode only cares whether mesh is active :D

PR #211, opened 2020-05-11
adjust injector logic for appmesh.k8s.aws/sidecarInjectorWebhook 

M00nF1sh:
1. adjust injector logic for `appmesh.k8s.aws/sidecarInjectorWebhook` as below:
   1. when `appmesh.k8s.aws/sidecarInjectorWebhook` presents on pod or namespace(pod takes priority):
        * `enabled`:  enforce sidecar injection, reject pod creation when VN absent.
        * `disabled`: skip sidecar injection
        *  other: inject when VN found, skip when not found.
   2. By default,   inject when VN found, skip when not found. (so the flag `enable-sidecar-injector` is no longer needed)
2. change metrics endpoint to 0.0.0.0:8080
3. adjust controller memory usage
4. remove metrics authProxy

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -28,7 +28,7 @@ patchesStrategicMerge:
   # Protect the /metrics endpoint by putting it behind auth.
   # If you want your controller-manager to expose the /metrics
   # endpoint w/o any authn/z, please comment the following line.
-- manager_auth_proxy_patch.yaml
+#- manager_auth_proxy_patch.yaml
 
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in 
 # crd/kustomization.yaml
--- a/main.go
+++ b/main.go
@@ -62,7 +62,7 @@ func init() {
 func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
-	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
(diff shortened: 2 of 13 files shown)


main.go
```diff
@@ -62,7 +62,7 @@ func init() {
 func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
-	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&metricsAddr, "metrics-addr", "0.0.0.0:8080", "The address the metric endpoint binds to.")
```
stefanprodan, line 65: We need to tell AppMesh Prometheus how to scrape the pods by adding this to the deployment spec:

```yaml
  template:
    metadata:
      annotations:
        prometheus.io/scrape: "true"
        prometheus.io/port: "8080"
```

config/default/kustomization.yaml
```diff
@@ -28,7 +28,7 @@ patchesStrategicMerge:
   # Protect the /metrics endpoint by putting it behind auth.
   # If you want your controller-manager to expose the /metrics
   # endpoint w/o any authn/z, please comment the following line.
-- manager_auth_proxy_patch.yaml
+#- manager_auth_proxy_patch.yaml
```
fawadkhaliq, line 31: Is this an intended change to disable authn/z ?

fawadkhaliq:
Overall changes look good

main.go
M00nF1sh: thanks :D adding :D

config/default/kustomization.yaml
M00nF1sh: yes, as most ppl don't have it setup :D

PR #216, opened 2020-05-12
make all injectors idempotent

M00nF1sh:
make all injectors logic idempotent :D so won't be issue even it's called twice :D 

### DIFF
--- a/pkg/inject/envoy.go
+++ b/pkg/inject/envoy.go
@@ -9,6 +9,7 @@ import (
 )
 
 const envoyTracingConfigVolumeName = "envoy-tracing-config"
+const envoyContainerName = "envoy"
 
 const envoyContainerTemplate = `
 {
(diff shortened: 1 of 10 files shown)


pkg/inject/envoy.go
```diff
@@ -166,3 +170,23 @@ func (m *envoyMutator) getPreview(pod *corev1.Pod) string {
 	}
 	return "0"
 }
+
+// containsEnvoyContainer checks whether pod already contains "envoy" container
+func containsEnvoyContainer(pod *corev1.Pod) bool {
+	for _, container := range pod.Spec.Containers {
+		if container.Name == envoyContainerName {
```
fawadkhaliq, line 177: Since this a name check, would this have an impact on envoy version bump workflows?

M00nF1sh: it don't. since pod spec can only be modified at pod creation time. :D

PR #227, opened 2020-05-13
add observedGeneration to all crds

M00nF1sh:
add observedGeneration to all crds.
This will help operator/tools to make sure configuration updates have been observed and proceeds by controller.


M00nF1sh:
> I don't believe this should be in the CRs themselves. This kind of information is germane to the _controller_, not the resource that the controller is managing. In other words, this belongs as state in the controller or resource manager, not the resource itself.
> 
> Coupling state like this into the resource itself means that the controller must ONLY be an active-passive architecture (i.e. only a single "writer" controller) and will preclude moving to an entirely stateless or multi-writer paradigm.
> 
> The controller struct should have a cache/hashmap of resource identifier (NamespacedName) to latest observed generation for the resource. And, frankly, that's what the client-go/cache utilities basically already are.

Hi jay, the information is for tools/operators to consume instead of controller itself.
Similar to deployment.status.Observedgeneration.

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md
> Some resources report the observedGeneration, which is the generation most recently observed by the component responsible for acting upon changes to the desired state of the resource. This can be used, for instance, to ensure that the reported status reflects the most recent desired status.

stefanprodan:
> Coupling state like this into the resource itself means that the controller must ONLY be an active-passive architecture (i.e. only a single "writer" controller) and will preclude moving to an entirely stateless or multi-writer paradigm.

@jaypipes  this is already the case, to guard against multiple active reconcilers, the controller has a leader election mechanism builtin.

@M00nF1sh I think is a great addition, Flagger and other controllers could verify that the desired state has been actuated. For example, Flagger could backoff on incrementing the traffic weight if the previous weight bump wasn't acknowledged by checking the observed generation (it does the same for Deployments and DaemonSets).

anguslees:
It's already been merged, and it's a minor thing but fwiw I agree with Jay: a single observedGeneration doesn't communicate anything useful.

You can't verify that the desired state has been actuated by this flag, only that the controller has seen it (and perhaps initiated some async action).  If you have a specific result you're looking for, *you need to look for that result* and not imply it from the value of observedGeneration and assume you know the implementation.

Jay's other point about multiple writers is also spot on - putting a single value in the schema like this (with the implied semantics of "when the desired state has been actuated") limits all possible future implementations to a single controller loop.  You can't (for example) have a coordinated effort from separate controller loops that all act on separate parts of the CRD and still capture that via a single observedGeneration.

Afaics observedGeneration is an obsolete concept that (quite rightly) is rarely implemented across the k8s ecosystem because it just doesn't provide useful semantics.

PR #230, opened 2020-05-16
Check for ns labels to determine sidecar injection

fawadkhaliq:
*Description of changes:*
To determine sidecar injection eligibility, we were matching on annotations instead of labels. We would see random injection misses when launching large number of pods at the same time. This was due to a race condition when VirtualNode would get created after it's corresponding pod.

This change introduces `namespaceSelector ` in the mutate webhook config to send label filtered requests to the injector

### DIFF
--- a/config/webhook/kustomization.yaml
+++ b/config/webhook/kustomization.yaml
@@ -4,3 +4,6 @@ resources:
 
 configurations:
 - kustomizeconfig.yaml
+
+patchesStrategicMerge:
+- pod_mutator_patch.yaml
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -120,7 +120,7 @@ webhooks:
       name: webhook-service
       namespace: system
       path: /mutate-v1-pod
-  failurePolicy: Ignore
+  failurePolicy: Fail
   name: mpod.appmesh.k8s.aws
   rules:
   - apiGroups:
--- a/config/webhook/pod_mutator_patch.yaml
+++ b/config/webhook/pod_mutator_patch.yaml
@@ -1,0 +1,9 @@
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: mutating-webhook-configuration
+webhooks:
+  - name: mpod.appmesh.k8s.aws
+    namespaceSelector:
+      matchLabels:
+        appmesh.k8s.aws/sidecarInjectorWebhook: enabled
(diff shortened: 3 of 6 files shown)


M00nF1sh:
lets do below instead to avoid accept every pod mutate request
1. remove the namespace related code here(and from UT as well)
2. add a file to `/config/webhook/pod_mutator_patch.yaml` (the kubebuilder mark don't support namespaceSelector yet)
```
apiVersion: admissionregistration.k8s.io/v1beta1
kind: MutatingWebhookConfiguration
metadata:
  name: mutating-webhook-configuration
webhooks:
  - name: mpod.appmesh.k8s.aws
    namespaceSelector:
      matchLabels:
```
3. modify `config/webhook/kustomization` to have additional content
```
patchesStrategicMerge:
- pod_mutator_patch.yaml
```

fawadkhaliq:
> lets do below instead to avoid accept every pod mutate request :D
> 
> 1. remove the namespace related code here(and from UT as well)

That's an excellent suggestion to limit the number of requests. I'll add the label match in the webhook config. However, on the removal part, wouldn't removing the namespace label check lead to silent failures again? If a VirtualNode is missing and the injector is not checking on namespace webhook enable/disable [1], it will ignore the injection [2] unless you meant we assume that everything that comes to the injector should inject the sidecar by default? If it's the latter, I think it'll be better to keep the `determineSidecarInjectMode` to check on namespace label. This will keep the injector self contained and will protect against any bugs/misconfig on the webhook config side

[1] https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/v1beta2/pkg/inject/inject.go#L136
[2] https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/v1beta2/pkg/inject/inject.go#L58

> 2. modify the annotation on "webhooks/core/pod_mutator" to be "failurePolicy=fail", given we'll have namespaceSelector below, it's make sense to enforce it since we already selected namespace by selector.
> 3. add a file to `/config/webhook/pod_mutator_patch.yaml` (the kubebuilder mark don't support namespaceSelector yet)
> 
> ```
> apiVersion: admissionregistration.k8s.io/v1beta1
> kind: MutatingWebhookConfiguration
> metadata:
>   name: mutating-webhook-configuration
> webhooks:
>   - name: mpod.appmesh.k8s.aws
>     namespaceSelector:
>       matchLabels:
>         appmesh.k8s.aws/sidecarInjectorWebhook: enabled
> ```
> 
> 1. modify `config/webhook/kustomization` to have additional content
> 
> ```
> patchesStrategicMerge:
> - pod_mutator_patch.yaml
> ```
 +1, I'll add the namespaceSelector based filter. Thanks!

M00nF1sh:
> unless you meant we assume that everything that comes to the injector should inject the sidecar by default?
Yeah, nice idea.  I think we can make the determineSidecarInjectMode be "enforce by default", unless user specify disable or other value (unspecified).
The check on namespace is redundant, since  its guaranteed namespace to have such label when webhook is called :D .

fawadkhaliq:
> remove the check leaves more flexibility, e.g. user can choose the label by modify webhook setting

That's a good point. I'll adjust this

PR #232, opened 2020-05-19
Allow overlapping virtual node names/labels across namespaces

fawadkhaliq:
*Description of changes:*
If two different namespaces have pods, virtual nodes with overlapping labels, the current controller would block creation of the pod as it would match against multiple virtual nodes. This not the desired behavior. We should be able to handle overlapping resource names, labels across namespaces. This change fixes the issue by filtering the virtual node list based on pod namespace


M00nF1sh:
awesome 👍

PR #241, opened 2020-05-26
Add Region and Availability zones as attributes to cloudmap instances

achevuru:
*Issue #, if available:*

*Description of changes:*
Add Region and Availability zones as attributes to cloudmap instances

### DIFF
--- a/pkg/cloudmap/instances_reconciler.go
+++ b/pkg/cloudmap/instances_reconciler.go
@@ -155,14 +155,12 @@ func (r *defaultInstancesReconciler) buildInstanceAttributes(ms *appmesh.Mesh, v
 	attr[attrK8sNamespace] = pod.Namespace
 	attr[attrAppMeshMesh] = aws.StringValue(ms.Spec.AWSName)
 	attr[attrAppMeshVirtualNode] = aws.StringValue(vn.Spec.AWSName)
-	if nodeInfoByName != nil {
-		if nodeInfo, ok := nodeInfoByName[podsNodeName]; ok {
-			if nodeInfo.region != "" {
-				attr[attrK8sPodRegion] = nodeInfo.region
-			}
-			if nodeInfo.availabilityZone != "" {
-				attr[attrK8sPodAZ] = nodeInfo.availabilityZone
-			}
+	if nodeInfo, ok := nodeInfoByName[podsNodeName]; ok {
+		if nodeInfo.region != "" {
+			attr[attrK8sPodRegion] = nodeInfo.region
+		}
+		if nodeInfo.availabilityZone != "" {
+			attr[attrK8sPodAZ] = nodeInfo.availabilityZone
 		}
 	}
 	return attr
--- a/pkg/cloudmap/instances_reconciler.go
+++ b/pkg/cloudmap/instances_reconciler.go
@@ -22,6 +22,10 @@ const (
 	attrK8sPod = "k8s.io/pod"
 	// AttrK8sNamespace is a custom attribute injected by app-mesh controller
 	attrK8sNamespace = "k8s.io/namespace"
+	// AttrK8sPodRegion is a custom attribute injected by app-mesh controller
+	attrK8sPodRegion = "REGION"
+	// AttrK8sPodAZ is a custom attribute injected by app-mesh controller
+	attrK8sPodAZ = "AVAILABILITY_ZONE"
 
 	attrAppMeshMesh        = "appmesh.k8s.aws/mesh"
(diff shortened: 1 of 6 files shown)


pkg/cloudmap/resource_manager.go
```diff
@@ -385,3 +390,32 @@ func (m *defaultResourceManager) isCloudMapServiceOwnedByVirtualNode(ctx context
 func (m *defaultResourceManager) buildCloudMapServiceSummaryCacheKey(nsSummary *servicediscovery.NamespaceSummary, serviceName string) string {
 	return fmt.Sprintf("%s/%s", awssdk.StringValue(nsSummary.Id), serviceName)
 }
+
+func (m *defaultResourceManager) getClusterNodeInfo(ctx context.Context) map[string]nodeAttributes {
```
fawadkhaliq, line 394: Is there a possibility of this going out of sync when a virtualNode is not touched but it's corresponding pod moves to a new node, where the new node is added after this function was called or the reconcile also takes into account Pod updates as well eliminating any possibility of out of sync?

achevuru: Yeah, CloudMap controller watches for pod events as well. So, we should be good with the above scenario..

pkg/cloudmap/instances_reconciler.go
```diff
@@ -135,19 +140,31 @@ func (r *defaultInstancesReconciler) buildInstanceInfoByID(ms *appmesh.Mesh, vn
...
 	for _, cmAttr := range vn.Spec.ServiceDiscovery.AWSCloudMap.Attributes {
 		attr[cmAttr.Key] = cmAttr.Value
 	}
+	podsNodeName := pod.Spec.NodeName
 	attr[attrAWSInstanceIPV4] = pod.Status.PodIP
 	attr[attrK8sPod] = pod.Name
 	attr[attrK8sNamespace] = pod.Namespace
 	attr[attrAppMeshMesh] = aws.StringValue(ms.Spec.AWSName)
 	attr[attrAppMeshVirtualNode] = aws.StringValue(vn.Spec.AWSName)
+	if nodeInfoByName != nil {
```
M00nF1sh, line 158: no need to do nil check for read-only map access :D

pkg/cloudmap/resource_manager.go
```diff
@@ -26,6 +27,9 @@ const (
 	defaultNamespaceCacheTTL               = 2 * time.Minute
 	defaultServiceCacheMaxSize             = 1024
 	defaultServiceCacheTTL                 = 2 * time.Minute
+
+	nodeRegionLabel           = "failure-domain.beta.kubernetes.io/region"
```
M00nF1sh, line 31: let's support "topology.kubernetes.io/region" and "topology.kubernetes.io/zone", these two annotation will replace the beta ones in v1.17 https://github.com/kubernetes/kubernetes/pull/81431

pkg/cloudmap/resource_manager.go
```diff
@@ -385,3 +390,32 @@ func (m *defaultResourceManager) isCloudMapServiceOwnedByVirtualNode(ctx context
...
 }
+
+func (m *defaultResourceManager) getClusterNodeInfo(ctx context.Context) map[string]nodeAttributes {
+	var nodeInfoByName map[string]nodeAttributes
+	nodeList := &corev1.NodeList{}
+	if err := m.k8sClient.List(ctx, nodeList); err != nil {
+		return nodeInfoByName
+	}
+
+	m.log.V(1).Info("Nodes ", "count: ", len(nodeList.Items))
```
M00nF1sh, line 401: better to be 
`m.log.V(1).Info("listed nodes", "count", len(nodeList.Items))`. don't use ":" :D

pkg/cloudmap/resource_manager.go
```diff
@@ -385,3 +390,32 @@ func (m *defaultResourceManager) isCloudMapServiceOwnedByVirtualNode(ctx context
 func (m *defaultResourceManager) buildCloudMapServiceSummaryCacheKey(nsSummary *servicediscovery.NamespaceSummary, serviceName string) string {
 	return fmt.Sprintf("%s/%s", awssdk.StringValue(nsSummary.Id), serviceName)
 }
+
+func (m *defaultResourceManager) getClusterNodeInfo(ctx context.Context) map[string]nodeAttributes {
+	var nodeInfoByName map[string]nodeAttributes
```
M00nF1sh, line 395: don't define nodeInfoByName here if it's not initialized later. we can just return nil if list fails.

PR #218, opened 2020-05-12
[v1beta2_ingress]: Add VirtualGateway Controller

karanvasnani:
**Issue #, if available:**
[aws-app-mesh-roadmap #37](https://github.com/aws/aws-app-mesh-roadmap/issues/37#issuecomment-614127518)

**Description of changes:**
Added virtualgateway controller.
The PR is split into following components:
- finalizer utility
- reference resolver
- types conversion from CRD to SDK
- resource manager
- controller & meshEvent handler
- auto-generated yaml code

**Testing:**
Since Ingress changes are still in preview and haven't been built into the aws-go-sdk, these changes cannot be tested with upstream aws-go-sdk. I built the go sdk locally by using the latest [preview model](https://raw.githubusercontent.com/aws/aws-app-mesh-roadmap/master/appmesh-preview/service-model.json) and used below command to run tests.
- make docker-build

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #248, opened 2020-06-03
Add resource manager and utils for virtualgateway

fawadkhaliq:
*Description of changes:*
Add resource manager and utils helpers for Virtual Gateway

### DIFF
--- a/pkg/virtualgateway/resource_manager.go
+++ b/pkg/virtualgateway/resource_manager.go
@@ -5,6 +5,7 @@ import (
 	appmesh "github.com/aws/aws-app-mesh-controller-for-k8s/apis/appmesh/v1beta2"
 	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/aws/services"
 	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/conversions"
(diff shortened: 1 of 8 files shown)


pkg/virtualgateway/resource_manager.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+func (m *defaultResourceManager) updateSDKVirtualGateway(ctx context.Context, sdkVG *appmeshsdk.VirtualGatewayData, ms *appmesh.Mesh, vg *appmesh.VirtualGateway) (*appmeshsdk.VirtualGatewayData, error) {
+	actualSDKVGSpec := sdkVG.Spec
+	desiredSDKVGSpec, err := m.buildSDKVirtualGatewaySpec(ctx, vg)
+	if err != nil {
+		return nil, err
+	}
+
+	opts := cmpopts.EquateEmpty()
```
M00nF1sh, line 160: see https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/218#discussion_r433386446  :D

fawadkhaliq: You're right, port is set by the server if left unset from the controller side. I'll fix the comp. Good catch 👍

PR #249, opened 2020-06-03
Add virtual gateway controller

fawadkhaliq:
*Description of changes:*

Controller for virtual gateway. This change adds:
- virtual gateway controller
- finalizers
- webhooks
- and enables the controller


pkg/virtualgateway/members_finalizer.go
```diff
@@ -0,0 +1,84 @@
...
+
+	message := m.buildPendingMembersEventMessage(ctx, grMembers)
+	m.eventRecorder.Eventf(vg, corev1.EventTypeWarning, "PendingMembersDeletion", message)
+	return runtime.NewRequeueAfterError(errors.New("pending members deletion"), m.evaluateInterval)
+}
+
+// findGatewayRouteMembers find the GatewayRoute members for this virtualGateway.
+func (m *pendingMembersFinalizer) findGatewayRouteMembers(ctx context.Context, vg *appmesh.VirtualGateway) ([]*appmesh.GatewayRoute, error) {
+	grList := &appmesh.GatewayRouteList{}
+	if err := m.k8sClient.List(ctx, grList); err != nil {
```
achevuru, line 62: Shouldn't this be namescoped?

pkg/virtualgateway/enqueue_requests_for_mesh_events.go
```diff
@@ -0,0 +1,71 @@
...
+	msOld := e.ObjectOld.(*appmesh.Mesh)
+	msNew := e.ObjectNew.(*appmesh.Mesh)
+
+	if mesh.IsMeshActive(msOld) != mesh.IsMeshActive(msNew) {
+		h.enqueueVirtualGatewaysForMesh(context.Background(), queue, msNew)
+	}
+}
+
+// Delete is called in response to a delete event
+func (h *enqueueRequestsForMeshEvents) Delete(e event.DeleteEvent, queue workqueue.RateLimitingInterface) {
```
achevuru, line 48: We would like to catch a mesh delete event as well isn't it? If Mesh referenced by this VirtualGateway is deleted, should we at least mark this VirtualGateway inactive?

pkg/virtualgateway/members_finalizer.go
fawadkhaliq: Yeah, that's good idea to make the list operation namespace scoped. It's filtered below right now based on vg. I'll adjust here and other places (mesh finalizers) in a subsequent improvement

pkg/virtualgateway/enqueue_requests_for_mesh_events.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// Update is called in response to an update event
+func (h *enqueueRequestsForMeshEvents) Update(e event.UpdateEvent, queue workqueue.RateLimitingInterface) {
+	// virtualGateway reconcile depends on mesh is active or not.
+	// so we only need to trigger virtualGateway reconcile if mesh's active status changed.
+	msOld := e.ObjectOld.(*appmesh.Mesh)
+	msNew := e.ObjectNew.(*appmesh.Mesh)
+
+	if mesh.IsMeshActive(msOld) != mesh.IsMeshActive(msNew) {
```
achevuru, line 42: Should we also check for a change in namespace selector for enqueuing the corresponding virtualgateway?

pkg/virtualgateway/enqueue_requests_for_mesh_events.go
fawadkhaliq: Good point, we should have finalizer in the mesh for virtual gateway

pkg/virtualgateway/enqueue_requests_for_mesh_events.go
fawadkhaliq: I think it should be fine as the virtual gateway still points to the same mesh

Although side note: it brings up an interesting point what happens when we try to point a mesh with active resources (virtualnodes etc) to a new namespace on the fly. Worth investigating to understand the behavior and manage accordingly. I've taken a note

PR #254, opened 2020-06-04
Add resource manager and utils for gatewayroute

fawadkhaliq:
*Description of changes:*
Add resource manager and util helpers for Gateway Route

### DIFF
--- a/pkg/gatewayroute/resource_manager.go
+++ b/pkg/gatewayroute/resource_manager.go
@@ -1,0 +1,357 @@
+package gatewayroute
+
+import (
+	"context"
+	appmesh "github.com/aws/aws-app-mesh-controller-for-k8s/apis/appmesh/v1beta2"
+	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/aws/services"
+	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/conversions"
+	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/k8s"
+	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/mesh"
+	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/references"
+	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/runtime"
+	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/virtualgateway"
+	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/virtualservice"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	appmeshsdk "github.com/aws/aws-sdk-go/service/appmesh"
+	"github.com/go-logr/logr"
+	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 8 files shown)


pkg/gatewayroute/resource_manager.go
```diff
@@ -0,0 +1,357 @@
...
+type defaultResourceManager struct {
+	k8sClient          client.Client
+	appMeshSDK         services.AppMesh
+	referencesResolver references.Resolver
+	accountID          string
+	log                logr.Logger
+}
+
+func (m *defaultResourceManager) Reconcile(ctx context.Context, gr *appmesh.GatewayRoute) error {
+	ms, err := m.findMeshDependency(ctx, gr)
```
achevuru, line 62: Is this necessary? wouldn't the below find + validate virtualGatewayDependency validate the Mesh as well? VirtualGateway wouldn't be up and active if the corresponding mesh is missing right?

pkg/gatewayroute/resource_manager.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+
+	return m.updateCRDGatewayRoute(ctx, gr, sdkGR)
+}
+
+func (m *defaultResourceManager) Cleanup(ctx context.Context, gr *appmesh.GatewayRoute) error {
+	ms, err := m.findMeshDependency(ctx, gr)
+	if err != nil {
+		return err
+	}
```
achevuru, line 107: So, if there is no virtualgateway associated with these set of gateway routes then can't we go ahead and delete it? Is there a reason we block the delete for a mesh being active?

pkg/gatewayroute/resource_manager.go
```diff
@@ -0,0 +1,357 @@
...
+
+func (m *defaultResourceManager) updateCRDGatewayRoute(ctx context.Context, gr *appmesh.GatewayRoute, sdkGR *appmeshsdk.GatewayRouteData) error {
+	oldGR := gr.DeepCopy()
+	needsUpdate := false
+	if aws.StringValue(gr.Status.GatewayRouteARN) != aws.StringValue(sdkGR.Metadata.Arn) {
+		gr.Status.GatewayRouteARN = sdkGR.Metadata.Arn
+		needsUpdate = true
+	}
+
+	if aws.Int64Value(gr.Status.ObservedGeneration) != gr.Generation {
```
achevuru, line 298: Should it be only when ObservedGeneration < Generation?

fawadkhaliq: Technically yes as this covers that too. Shouldn't cause any issues either way. Do you see any issues with the current approach?

pkg/gatewayroute/resource_manager.go
achevuru: No, my question is why do we need to check if there is an active mesh in the namespace before deleting gateway routes? Shouldn't we just check for virtual gateway? I can just choose to remove VG and GR and should be able to continue to use the mesh like how we do currently, right?

pkg/gatewayroute/resource_manager.go
achevuru: I mean, once observedGeneration becomes > generation not sure how they can ever catch up. We will end up updating the resource every time even when there is no need.

Issue #75, opened 2019-09-22
Support App Mesh Preview Channel


bcelenza:
Allow the controller to be used for the [App Mesh Preview Channel](https://docs.aws.amazon.com/app-mesh/latest/userguide/preview.html), so new features can be tested more easily with Kubernetes.

Currently this package relies solely on the published SDK Go client, which will not contain all of the features in preview.

bcelenza:
With the merge of #257 the controller is capable of running against preview. Next step before closing this out is to produce a managed Docker image for preview, the same as we do for production.

fawadkhaliq:
We will be releasing managed Docker images for preview from now. This is already done for v1.0.0-preview. e.g `602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon/appmesh-controller:v1.0.0-preview`. This issue is good to close

PR #257, opened 2020-06-05
Generate SDK from published preview models

bcelenza:
*Issue #, if available:*
aws/aws-app-mesh-controller-for-k8s#75

*Description of changes:*

Changes the Go SDK generation to use the [published SDK models for preview](https://github.com/aws/aws-app-mesh-roadmap/pull/210). This will allow us to support the controller working with the preview environment long term.

One minor side effect was needing to delete the `Tags: nil` references throughout the resource managers, as we don't publish the tagging operations in the preview SDK (we don't yet support it in preview). If and when we want tag support in the controller, we can definitely add it to the preview SDK -- but for now it isn't necessary to explicitly send the nil values.

*Testing:*

Built a regular and preview version of the controller:

```
IMG=bcelenza/app-mesh-controller:latest make docker-build
APPMESH_PREVIEW=1 IMG=bcelenza/app-mesh-controller:latest-preview make docker-build
```
Pushed the resulting images to docker hub.

Deployed the regular image to my IAD cluster to talk to the prod app mesh endpoint

Deployed the preview version to my preview cluster in PDX.

Tested creating a mesh in both clusters and verified they were created against the appropriate app mesh endpoints.


appmesh_models_override/setup.sh
```diff
@@ -4,16 +4,33 @@ set -e
 
 mkdir -p ./vendor/github.com/aws
 
+SDK_MODEL_SOURCE=https://raw.githubusercontent.com/aws/aws-app-mesh-roadmap/master/appmesh-preview/sdk
 SDK_VENDOR_PATH=./vendor/github.com/aws/aws-sdk-go
+API_VERSION=2019-01-25
+API_PATH=$SDK_VENDOR_PATH/models/apis/appmesh/$API_VERSION
+SERVICE_NAME=$([ "$APPMESH_PREVIEW" == "1" ] && echo "appmesh-preview" || echo "appmesh" )
```
fawadkhaliq, line 11: Does not really impact this CR as this was the case before as well. Will `appmesh` service endpoint support `appmesh-preview` source model at line#7 or we expect some features may not be supported?

fawadkhaliq:
Thanks for adding this
/lgtm

appmesh_models_override/setup.sh
bcelenza: The App Mesh production endpoint will not support all features in preview, but preview will support all features from App Mesh production. In effect, the preview SDK long-term will be a superset of production.

For features not supported in production, the controller will either error with a 404 (API doesn't exist) or error with a 403 (access denied). We protect new features with feature flags until they're ready to go GA, so it's generally safe to vend the preview models for production use.

fawadkhaliq: Perfect, makes sense. Thanks!

PR #262, opened 2020-06-09
Add config mutator for virtual gateway envoy

fawadkhaliq:
*Description of changes:*
Add config mutator for virtual gateway envoy.

If a virtual gateway podSelecter matches on a pod that is not marked for virtual node already, it will inject virtual gateway config to it. The mutated config is added as `env` for the pod and it includes reference to the virtualGateway resources in App Mesh

### DIFF
--- a/pkg/inject/virtualgateway_envoy.go
+++ b/pkg/inject/virtualgateway_envoy.go
@@ -1,0 +1,121 @@
+package inject
+
+import (
+	"encoding/json"
+	appmesh "github.com/aws/aws-app-mesh-controller-for-k8s/apis/appmesh/v1beta2"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	"strings"
+)
+
+const envoyImageStub = "injector-envoy-image"
+const envoyVirtualGatewayEnvMap = `
+{
+  "APPMESH_VIRTUAL_NODE_NAME": "mesh/{{ .MeshName }}/virtualGateway/{{ .VirtualGatewayName }}",
+  "APPMESH_PREVIEW": "{{ .Preview }}",
+  "ENVOY_LOG_LEVEL": "{{ .LogLevel }}",
+  "AWS_REGION": "{{ .AWSRegion }}"
+}
+`
+
+type VirtualGatewayEnvoyVariables struct {
+	AWSRegion          string
+	MeshName           string
+	VirtualGatewayName string
+	Preview            string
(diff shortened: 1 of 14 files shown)


pkg/inject/virtualgateway_envoy.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return err
+	}
+
+	//we override the image to latest Envoy so customers do not have to manually manage
+	// envoy versions and let controller handle consistency versions across the mesh
+	if pod.Spec.Containers[envoyIdx].Image == envoyImageStub || pod.Spec.Containers[envoyIdx].Image == m.mutatorConfig.sidecarImage {
+		pod.Spec.Containers[envoyIdx].Image = m.mutatorConfig.sidecarImage
+	} else {
+		return errors.Errorf("invalid envoy image name for injection %s, expected name: %s", pod.Spec.Containers[envoyIdx].Image, envoyImageStub)
```
bcelenza, line 78: I didn't think to mention this before (sorry!), but we will want to reserve the possibility that someone is bringing their own Envoy image for the gateway as well. Would it make more sense to just mutate if `name==envoy &&  image==injector-envoy-image`, otherwise allowing the provided configuration?

fawadkhaliq: I thought about it. The downside is that it also opens the possibility of customer programming invalid images like not-envoy or envoy-image-with-missing-features or envoy-image-not-longer-supported. If we're okay with the risk, we can open it up. What do you think?

fawadkhaliq: Also, in either case, it's definitely a two-way door decision. And we do currently support overriding sidecar image [1]

[1] https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/v1beta2/pkg/inject/config.go#L76

bcelenza: Yeah, I think there’s a rare case where the gateway needs to be a custom envoy while the sidecars may be something else. I think most customers will prefer consistency, but allowing them to bring their own image is something we’ve always wanted to provide as an option (even though we may not be able to fully support it).

fawadkhaliq: Sounds good. We can allow that. I will push a subsequent PR after this to allow for custom images

fawadkhaliq: Added the option to side override https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/266. Added an annotation instead of using a special image to determine the condition

PR #268, opened 2020-06-11
Change annotation value from true to enabled

fawadkhaliq:
*Description of changes:*
Two of the annotations used "true" as a value. This would require extra quotes in the YAML and unquoted value would mess up the YAML. This change uses a better name for annotation values `appmesh.k8s.aws/preview` and `appmesh.k8s.aws/virtualGatewaySkipImageOverride`.

After this change, the annotations for App Mesh preview and Virtual Gateway skip image override will be:

```
annotations:
  appmesh.k8s.aws/preview: enabled
  appmesh.k8s.aws/virtualGatewaySkipImageOverride: enabled
```

### DIFF
--- a/pkg/inject/virtualgateway_envoy.go
+++ b/pkg/inject/virtualgateway_envoy.go
@@ -142,9 +142,9 @@ func (m *virtualGatewayEnvoyConfig) virtualGatewayImageOverride(pod *corev1.Pod)
 	}
 
 	switch strings.ToLower(imageOverrideAnnotation) {
-	case "enabled":
+	case gatewayImageOverrideModeEnabled:
 		return false
-	case "disabled":
+	case gatewayImageOverrideModeDisabled:
 		return true
 	default:
 		return true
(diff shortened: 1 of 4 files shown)


pkg/inject/virtualgateway_envoy.go
```diff
@@ -127,12 +127,27 @@ func (m *virtualGatewayEnvoyConfig) getAugmentedMeshName() string {
 	return meshName
 }
 
+const (
+	// when enabled, a virtual gateway image will not be overriden
+	gatewayImageOverrideModeEnabled = "enabled"
```
M00nF1sh, line 132: humm, seems this two const is not used..

fawadkhaliq: oops, my bad. A missed change. Fixed it. Thanks for catching it

M00nF1sh: this is a little counter-intutive, maybe GatewaySkipImageOverrideMode?

fawadkhaliq: dang - of course 🤦  😄

PR #273, opened 2020-06-12
Extract object namespace information from object request context

fawadkhaliq:
*Description of changes:*
Namespace of submitted objects is not trustworthy in Mutating Webhooks, and may be blank in Validation Webhooks. The namespace needs to always be read out of the AdmissionReview object. See https://github.com/kubernetes/kubernetes/issues/88282

Changed the VirtualNode and VirtualGateway designators to extract namespace information from the admission attributes in the create request

### DIFF
--- a/pkg/virtualnode/membership_designator.go
+++ b/pkg/virtualnode/membership_designator.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	appmesh "github.com/aws/aws-app-mesh-controller-for-k8s/apis/appmesh/v1beta2"
 	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/k8s"
+	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/webhook"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


pkg/virtualnode/membership_designator.go
```diff
@@ -33,8 +35,15 @@ type membershipDesignator struct {
 // +kubebuilder:rbac:groups=appmesh.k8s.aws,resources=virtualnodes,verbs=get;list;watch
 
 func (d *membershipDesignator) Designate(ctx context.Context, pod *corev1.Pod) (*appmesh.VirtualNode, error) {
+
+	// see https://github.com/kubernetes/kubernetes/issues/88282 and https://github.com/kubernetes/kubernetes/issues/76680
+	req := webhook.ContextGetAdmissionRequest(ctx)
+	podNS := &corev1.Namespace{}
+	if err := d.k8sClient.Get(ctx, types.NamespacedName{Name: req.Namespace}, podNS); err != nil {
```
M00nF1sh, line 42: i don't think we need this namespace exists check here :D

fawadkhaliq: good catch! optimized

PR #275, opened 2020-06-16
Inject xray daemon to virtual gateway pod

fawadkhaliq:
*Description of changes:*
Inject xray daemon to virtual gateway pod

### DIFF
--- a/pkg/inject/virtualgateway_envoy.go
+++ b/pkg/inject/virtualgateway_envoy.go
@@ -15,7 +15,8 @@ const envoyVirtualGatewayEnvMap = `
   "APPMESH_VIRTUAL_NODE_NAME": "mesh/{{ .MeshName }}/virtualGateway/{{ .VirtualGatewayName }}",
   "APPMESH_PREVIEW": "{{ .Preview }}",
   "ENVOY_LOG_LEVEL": "{{ .LogLevel }}",
-  "AWS_REGION": "{{ .AWSRegion }}"
+  "AWS_REGION": "{{ .AWSRegion }}"{{ if .EnableXrayTracing }},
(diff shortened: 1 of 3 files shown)


pkg/inject/virtualgateway_envoy.go
```diff
@@ -15,7 +15,8 @@ const envoyVirtualGatewayEnvMap = `
   "APPMESH_VIRTUAL_NODE_NAME": "mesh/{{ .MeshName }}/virtualGateway/{{ .VirtualGatewayName }}",
   "APPMESH_PREVIEW": "{{ .Preview }}",
   "ENVOY_LOG_LEVEL": "{{ .LogLevel }}",
-  "AWS_REGION": "{{ .AWSRegion }}"
+  "AWS_REGION": "{{ .AWSRegion }}"{{ if .EnableXrayTracing }},
+  "ENABLE_ENVOY_XRAY_TRACING": "1"{{ end }}
```
mogren, line 19: Tricky with the if templating here, haven't seen that before. 😄

fawadkhaliq:
> Looks a lot like the Xray mutator for virtual node... 😀

That's correct. It's the same [xray mutator](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/inject/xray.go) that is enabled here for virtual gateways

pkg/inject/virtualgateway_envoy.go
fawadkhaliq: +1, I agree but going with the existing [model](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/inject/envoy.go#L30) for now. Can definitely be done differently

PR #276, opened 2020-06-16
Rename manager to controller for consistency with docs

fawadkhaliq:
*Description of changes:*
We have been using manager and controller interchangeably for the App Mesh controller. This causes confusion in the documents. Renaming manager to controller to keep it consistent across the board


fawadkhaliq:
> /awesome
> have we verified the generated deploy.yaml don't contain "manager"?

Yes, end to end verification done: yaml scanning and mesh with workloads

PR #279, opened 2020-06-17
Update make targets to publish images to ECR

fawadkhaliq:
*Description of changes:*
Update make targets to publish images to ECR

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 # Image URL to use all building/pushing image targets
-IMAGE=amazon/appmesh-controller
-REPO=$(AWS_ACCOUNT).dkr.ecr.$(AWS_REGION).amazonaws.com/$(IMAGE)
-VERSION ?= latest
-DEV_VERSION=$(shell git describe --dirty --tags)
+IMAGE_NAME=amazon/appmesh-controller
+REPO=$(AWS_ACCOUNT).dkr.ecr.$(AWS_REGION).amazonaws.com/$(IMAGE_NAME)
+VERSION ?= $(shell git describe --dirty --tags)
+IMAGE ?= $(REPO):$(VERSION)
 
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -39,7 +42,7 @@ uninstall: manifests
 
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
 deploy: manifests
-	cd config/controller && kustomize edit set image controller=${IMG}
+	cd config/controller && kustomize edit set image controller=${IMAGE}:${DEV_VERSION}
```
M00nF1sh, line 45: i believe it should be ${REPO}:${DEV_VERSION}

Makefile
```diff
@@ -60,11 +63,19 @@ generate: controller-gen
 
 # Build the docker image
 docker-build: test
-	docker build . -t ${IMG}
+	docker build . -t ${IMAGE}:${DEV_VERSION}
+
+# Build the release docker image
+docker-build-release: test
```
M00nF1sh, line 69: 1. why don't build it to be REPO directly instead of retag it.
2. i don't think we should use 'latest" version here. it should be like v1.0.0

M00nF1sh: a better logic might be 
1. avoid this build-release and push-release(instead only have this original docker-build & docker-push command).
2. specify image to be (AWS_ACCOUNT).dkr.ecr.$(AWS_REGION).amazonaws.com/amazon/appmesh-controller
3. use a single version instead of version and devVersion, which is result of "git describe --tags --dirty --always"
4. if we want to release a GA version, we just tag the commit like v1.0.0, and docker-build && docker-push
5. if we want to release a Dev version, we just override the IMAGE completely or just aws account.

achevuru: +1 on avoiding build-release & push-release and instead using existing docker-build and docker-push. I believe we had this build-release, push-release format in the old controller...

fawadkhaliq: I'm more inclined towards having explicit targets for whats being done rather than relying on the caller but both approaches are fine with the use of env. I will adjust this

Makefile
fawadkhaliq: Done

PR #284, opened 2020-06-18
Update the README note about older versions and injector

fawadkhaliq:
*Description of changes:*
Update the README note about older versions and injector


fawadkhaliq:
> Is this terminology change "AWS App Mesh Controller For Kubernetes" from "AWS App Mesh Controller For K8s" consistent with other places in the documentation ?

That's a good point. Let's keep it k8s to avoid any inconsistency if people are searching using these. In any case, shouldn't be a problem as k8s and Kubernetes are used interchangeably

Issue #288, opened 2020-02-12
Add Envoy preStop hook


quentinplessis:
When a pod is terminating there is a time lag before the pod's removal is propagated to all envoys of the mesh. During this time, requests may still be sent to the terminating pod.
Adding a preStop hook to envoy to wait for the pod's removal to be propagated before shutting down would be a way to solve this problem.

achevuru:
Added support for configurable preStop delay for envoy containers via #312

PR #312, opened 2020-07-13
Envoy PreStop hook support

achevuru:
*Issue #, if available:*
#288
*Description of changes:*

Support for a configurable pre-stop delay for Envoy containers. 

- Will help with keeping envoy up and running until the app container goes down. Sidecar feature in K8S 1.18 will address this natively.
- Delay also helps to account for the time it takes for the pod down event to propagate to the remaining envoys in the Mesh.
- Default value of 20 sec is based on some internal tests and specifically tuned for CloudMap based ServiceDiscovery mode and can be adjusted higher based on the scale. Value can be adjusted lower for DNS scenarios.
- If preStop delay is set to a value greater than 30 secs, then terminationGracePeriodSeconds in the pod spec (spec.terminationGracePeriodSeconds) has to be adjusted accordingly. If not adjusted, Pod will be killed(via SIGKILL) prior to the completion of envoy preStop hook as the default value of terminationGracePeriodSeconds is 30.


fawadkhaliq:
> If preStop delay is set to a value greater than 30 secs, then terminationGracePeriodSeconds has to be adjusted accordingly

Looks good overall. In the description, can we please elaborate more on the 30 seconds with specifics. Why and where `terminationGracePeriodSeconds` needs to be changed. A sample pod spec would be great! Will be helpful for the users and anyone updating documentation

PR #324, opened 2020-07-31
Support for Timeout Integration tests

achevuru:
*Issue #, if available:*

N/A

*Description of changes:*

Support for Timeout Integration tests. Adds a test suite to verify appmesh's timeout feature.
 - Verifies that the request times out if it takes more than the default 15s without timeout configured.
 - Verifies that the same request goes  through fine with appropriate timeout value configured.

Test Output:

STEP: verify route timesout if it takes more than default 15s w/o listener timeout configured
{"level":"warn","ts":1596217978.377368,"msg":"Received Response: ","Response: ":"upstream request timeout","StatusCode: ":200}
STEP: verify the timeout behaviour with configured timeout value
{"level":"warn","ts":1596218023.717762,"msg":"Received Response: ","Response: ":"backend","StatusCode: ":200}

### DIFF
--- a/test/integration/timeout/timeout_stack.go
+++ b/test/integration/timeout/timeout_stack.go
@@ -25,8 +25,10 @@ import (
 )
 
 const (
-	defaultFrontEndImage = "928111597794.dkr.ecr.us-west-2.amazonaws.com/amazon/timeout-e2e:feapp-final"
-	defaultBackEndImage  = "928111597794.dkr.ecr.us-west-2.amazonaws.com/amazon/timeout-e2e:colorapp-final-1"
+	//If you're not able to access below images, try to build them based on the app code under "timeout_app"
+	//directory and push it to any accessible ECR repo and update the below values
+	defaultFrontEndImage = "928111597794.dkr.ecr.us-west-2.amazonaws.com/amazon/timeout-e2e:frontend-image"
+	defaultBackEndImage  = "928111597794.dkr.ecr.us-west-2.amazonaws.com/amazon/timeout-e2e:backend-image"
 
 	timeoutTest      = "timeout-e2e"
 	AppContainerPort = 8080
@@ -37,14 +39,14 @@ const (
 
 // Timeout stack is setup as below
 //	FrontEnd ->
-//        - Exposes two endpoints "/default" and "/timeout" and reaches out to backend for both the endpoints
+//        - Exposes two endpoints "/defaultroute" and "/timeoutroute" and reaches out to backend for both the endpoints
 //        - Refers to "backend" VirtualService which in turn refers to "backend" VirtualRouter
 //
 //  Backend ->
-//        - Exposes "/default" and "/timeout" with a configured 45 seconds delay
+//        - Exposes "/defaultroute" and "/timeoutroute" with a configured 45 seconds delay
 //        - "backend" VN is configured with a Listener timeout of 60 seconds
 //        - "backend" VR has two routes "/default" and "/timeout"
-//        - "/default" path uses default timeout of 15s and "/timeout" path is configured with 60s timeout
+//        - "/defaultroute" path uses default timeout of 15s and "/timeoutroute" path is configured with 60s timeout
 //
 //  We then validate the timeout feature.
 //       - Call to "/default" from frontend pod should timeout with "upstream request timeout"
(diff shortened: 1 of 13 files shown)


test/integration/timeout/timeout_stack_test.go
```diff
@@ -0,0 +1,60 @@
...
+				It(fmt.Sprintf("Should behave correctly with and without timeout configured"), func() {
+					stackPrototype.ServiceDiscoveryType = sdType
+					stackDefault := stackPrototype
+
+					By("deploy timeout stack into cluster", func() {
+						stacksPendingCleanUp = append(stacksPendingCleanUp, &stackDefault)
+						stackDefault.DeployTimeoutStack(ctx, f)
+					})
+
+					By("sleep 30 seconds for Envoys to be configured", func() {
```
fawadkhaliq, line 49: Can we please add a TODO here to remove this once readiness probe PR https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/325 is merged

test/integration/timeout/timeout_stack.go
```diff
@@ -0,0 +1,666 @@
...
+// Timeout stack is setup as below
+//	FrontEnd ->
+//        - Exposes two endpoints "/default" and "/timeout" and reaches out to backend for both the endpoints
+//        - Refers to "backend" VirtualService which in turn refers to "backend" VirtualRouter
+//
+//  Backend ->
+//        - Exposes "/default" and "/timeout" with a configured 45 seconds delay
+//        - "backend" VN is configured with a Listener timeout of 60 seconds
+//        - "backend" VR has two routes "/default" and "/timeout"
+//        - "/default" path uses default timeout of 15s and "/timeout" path is configured with 60s timeout
```
fawadkhaliq, line 47: The description of the stack is very helpful. Thanks for adding it!

The test relies of hosted images and part of the logic relies on the application. Can we please:
1) commit the application code in this suite so it's all contained here. Don't have to add build->publish->run in this suite support yet - we can add that later
2) remove the fixed time out from the application code and make the time out values picked up as part of env variables in the application that you can define as part of Pod spec

test/integration/timeout/timeout_stack.go
```diff
@@ -0,0 +1,666 @@
...
+					zap.Error(err),
+				)
+				deletionErrors = append(deletionErrors, err)
+			}
+		})
+	})
+	return deletionErrors
+}
+
+func (s *TimeoutStack) checkDefaultBackendRouteTimesout(ctx context.Context, f *framework.Framework,
```
fawadkhaliq, line 584: `checkDefaultBackendRouteTimesout` and `checkBehaviorOfRouteConfiguredWithTimeout` are very similar except for the one line of error message check. You could merge them into one function and have a check based on a condition

test/integration/timeout/timeout_stack.go
```diff
@@ -0,0 +1,666 @@
...
+	}
+	for _, err := range deletionErrors {
+		f.Logger.Error("clean up failed", zap.Error(err))
+	}
+	Expect(len(deletionErrors)).To(BeZero())
+}
+
+//Check Timeout behavior with and with timeout configured
+func (s *TimeoutStack) CheckTimeoutBehavior(ctx context.Context, f *framework.Framework) {
+	By(fmt.Sprintf("verify route timesout if it takes more than default 15s w/o listener timeout configured"), func() {
```
fawadkhaliq, line 107: Would be nice to make the 15 second configurable and see if Envoy sees the change - maybe enhance it later

test/integration/timeout/timeout_stack.go
```diff
@@ -0,0 +1,666 @@
...
+
+// Timeout stack is setup as below
+//	FrontEnd ->
+//        - Exposes two endpoints "/default" and "/timeout" and reaches out to backend for both the endpoints
+//        - Refers to "backend" VirtualService which in turn refers to "backend" VirtualRouter
+//
+//  Backend ->
+//        - Exposes "/default" and "/timeout" with a configured 45 seconds delay
+//        - "backend" VN is configured with a Listener timeout of 60 seconds
+//        - "backend" VR has two routes "/default" and "/timeout"
```
fawadkhaliq, line 46: nit: better to use different names for virtualrouter paths something like `/rtr-default ` and `/rtr-timeout `. When I was reviewing the http.get operation, I wasn't sure if the call is intended for application directly or through virtualrouter in the topology - can be confusing

test/integration/timeout/timeout_stack_test.go
achevuru: Readiness probe might not entirely take care of it here right?

fawadkhaliq: It can help make it more deterministic. Once the readiness probe is there and if Envoy is not ready, the container and pod condition `Ready` and `ContainersReady` will be set to false. You could query that (with a max # of retries) and move forward when pod is ready

Here's a sample of status->conditions for a pod which has readiness probe failing:
```
        "conditions": [
            {
                "lastProbeTime": null,
                "lastTransitionTime": "2020-07-31T22:10:55Z",
                "status": "True",
                "type": "Initialized"
            },
            {
                "lastProbeTime": null,
                "lastTransitionTime": "2020-07-31T22:10:55Z",
                "message": "containers with unready status: [envoy]",
                "reason": "ContainersNotReady",
                "status": "False",
                "type": "Ready"
            },
            {
                "lastProbeTime": null,
                "lastTransitionTime": "2020-07-31T22:10:55Z",
                "message": "containers with unready status: [envoy]",
                "reason": "ContainersNotReady",
                "status": "False",
                "type": "ContainersReady"
            },
            {
                "lastProbeTime": null,
                "lastTransitionTime": "2020-07-31T22:10:55Z",
                "status": "True",
                "type": "PodScheduled"
            }
        ],
```

achevuru: No, I mean we are not waiting here for Envoy to be ready and responsive here but more for the config to be populated from appmesh control plane. We can discuss..

Issue #289, opened 2019-06-11
Add optional limits to injected Containers


norwoodj:
We deploy ResourceQuotas in each of our application namespaces, in order to force teams to put memory/CPU requests on all of their pods. As a result, our injected pods cannot start, because the `proxyinit` initContainer does not specify any resources requests or limits. We request that the init container has resources and limit specified.

M00nF1sh:
we should expose configurability for cpu/memory limits as well.(in addition to requests)

fawadkhaliq:
Optional resource limits for init and sidecar containers added here #326

PR #326, opened 2020-08-03
Add optional resource limits for init and sidecar containers

fawadkhaliq:
*Issue* #289  

*Description of changes:*
Add optional cpu, memory limits for init and sidecar containers.

Resource limits are opt-in for App Mesh users and are *disabled* by default. There are two ways to set resource limits:
1. Cluster wide: by setting `sidecar-cpu-limits` and/or `sidecar-memory-limits` flags on the controller
2. Per pod: by setting pod annotations `appmesh.k8s.aws/cpuLimit` and/or `appmesh.k8s.aws/memoryLimit`. Per pod annotation will override the cluster setting if both are specified.

### DIFF
--- a/pkg/inject/init_proxy.go
+++ b/pkg/inject/init_proxy.go
@@ -46,13 +46,7 @@ const proxyInitContainerTemplate = `
       "name": "APPMESH_EGRESS_IGNORED_PORTS",
       "value": "{{ .EgressIgnoredPorts }}"
     }
-  ],
-  "resources": {
-    "requests": {
-      "cpu": "{{ .CPURequests }}",
-      "memory": "{{ .MemoryRequests }}"
-    }
-  }
+  ]
 }
 `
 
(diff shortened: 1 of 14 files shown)


pkg/inject/init_proxy.go
```diff
@@ -102,6 +96,14 @@ func (m *initProxyMutator) mutate(pod *corev1.Pod) error {
 	if err != nil {
 		return err
 	}
+
+	// add resource requests and limits
+	container.Resources, err = sidecarResources(m.mutatorConfig.cpuRequests, m.mutatorConfig.memoryRequests,
```
achevuru, line 101: Why are we not checking for pod annotations here for resource limits?

pkg/inject/xray_test.go
```diff
@@ -159,6 +192,70 @@ func Test_xrayMutator_mutate(t *testing.T) {
 									"cpu":    cpuRequests,
 									"memory": memoryRequests,
 								},
+								Limits: corev1.ResourceList{
```
achevuru, line 195: Would be good to add the same test case scenarios that are added for envoy around this.

pkg/inject/init_proxy.go
fawadkhaliq: init container does not really depend on your application datapath load. Init container here will run as one shot, do it's job and finish execution. Whereas envoy or xray may need higher or lower resource limits depending on your application as they come in the datapath

pkg/inject/xray_test.go
fawadkhaliq: Added

pkg/inject/init_proxy.go
fawadkhaliq: There might be a bit of user experience aspect that users would need to know how to enable limits for init containers. We will document it

PR #327, opened 2020-08-03
Add integration tests for virtual service

fawadkhaliq:
*Description of changes:*
Add integration tests for virtual service custom resource. Covers basic create, update and delete scenarios

### DIFF
--- a/test/framework/resource/virtualservice/manager.go
+++ b/test/framework/resource/virtualservice/manager.go
@@ -2,7 +2,6 @@ package virtualservice
 
 import (
 	"context"
-	"time"
 
 	appmesh "github.com/aws/aws-app-mesh-controller-for-k8s/apis/appmesh/v1beta2"
 	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/aws/services"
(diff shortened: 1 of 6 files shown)


test/integration/virtualservice/virtualservice.go
```diff
@@ -0,0 +1,133 @@
...
+
+	for _, err := range deletionErrors {
+		f.Logger.Error("VirtualService clean up failed", zap.Error(err))
+	}
+	Expect(len(deletionErrors)).To(BeZero())
+}
+
+func (m *VirtualServiceTest) CheckInAWS(ctx context.Context, f *framework.Framework, ms *appmesh.Mesh, vs *appmesh.VirtualService) error {
+	var err error
+	// Reconcile may take a while so add re-tries
```
achevuru, line 120: Thinking about it, based on our experience from the scale tests the current value of 1 sec might not hold if we scale no.of Virtualservices. How about we adjust "retryWait" based on the scale? If not, we should introduce something in the lines of "initialDelay" that is configurable based on the scale in the test suite before we land here.

test/framework/resource/virtualservice/manager.go
```diff
@@ -40,8 +42,18 @@ type defaultManager struct {
 func (m *defaultManager) WaitUntilVirtualServiceActive(ctx context.Context, vs *appmesh.VirtualService) (*appmesh.VirtualService, error) {
 	observedVS := &appmesh.VirtualService{}
 	return observedVS, wait.PollImmediateUntil(utils.PollIntervalShort, func() (bool, error) {
-		if err := m.k8sClient.Get(ctx, k8s.NamespacedName(vs), observedVS); err != nil {
-			return false, err
+
+		// sometimes there's a delay in the resource showing up
```
achevuru, line 46: I remember we did something similar for VN and VR but we shouldn't need this change isn't it? "wait.PollImmediateUntil" should keep on trying the func until it returns true.I don't see a reason for this change, what am I missing?

fawadkhaliq: `PollImmediateUntil` tries the func until it returns true or error so in this case `Get` returns an err that virtualservice is not found. So we need an additional check to re-try on error because of eventual consistency. I'll adjust this to align better so `PollImmediateUntil` can retry instead of doing it within the func

test/integration/virtualservice/virtualservice.go
fawadkhaliq: Initial delay isn't really deterministic so not a fan of that approach :( 

You're right on the scale. Since these aren't scale tests right now, we can adjust the retry wait when we adapt these for scale

Issue #295, opened 2020-06-18
[Enhancement] Add more client-side validation via our validatingWebhook


M00nF1sh:
We should add more client-side validation via our validatingWebhook

fawadkhaliq:
Client side validation for open issues (https://github.com/aws/aws-app-mesh-controller-for-k8s/issues/135) has been added here #331. In this change:

- VirtualNode validator will check for duplicate backend references (VirtualServiceRefs & VirtualServiceARNs) and reject them on create/update
- VirtualRouter validator will check for duplicate route entries and reject them on create/update

PR #331, opened 2020-08-04
Enhance Validation for VirtualNodes and VirtualRouters specs

achevuru:
*Issue #, if available:*
Enhance Validation for VirtualNodes and VirtualRouters specs - #295

*Description of changes:*

- VirtualNode specs will now be validated/checked for the presence for duplicate backend references (VirtualServiceRefs & VirtualServiceARNs)
- VirtualRouter specs will now be validated/checked for the presence for duplicate route entries

### DIFF
--- a/webhooks/appmesh/virtualnode_validator.go
+++ b/webhooks/appmesh/virtualnode_validator.go
@@ -3,6 +3,7 @@ package appmesh
 import (
 	"context"
 	appmesh "github.com/aws/aws-app-mesh-controller-for-k8s/apis/appmesh/v1beta2"
+	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/references"
 	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/webhook"
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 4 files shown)


webhooks/appmesh/virtualnode_validator.go
```diff
@@ -63,6 +71,30 @@ func (v *virtualNodeValidator) enforceFieldsImmutability(vn *appmesh.VirtualNode
...
+func (v *virtualNodeValidator) checkVirtualNodeBackendsForDuplicates(vn *appmesh.VirtualNode) error {
+	backends := vn.Spec.Backends
+	backendMap := make(map[string]bool, len(backends))
+
+	for _, backend := range backends {
+		if backend.VirtualService.VirtualServiceRef != nil {
+			backendNamespacedName := references.ObjectKeyForVirtualServiceReference(vn, *backend.VirtualService.VirtualServiceRef)
+			backendIdentifier := backendNamespacedName.Name + "-" + backendNamespacedName.Namespace
+			if _, ok := backendMap[backendIdentifier]; ok {
+				return errors.Errorf("%s-%s has duplicate VirtualServiceReferences %s", "VirtualNode", vn.Name, backend.VirtualService.VirtualServiceRef.Name)
```
fawadkhaliq, line 83: nit: better to log the namespace name as well. Will help with easier debugging in large clusters with duplicate names across namespaces

achevuru: Name of the duplicate service should help us find out the duplicates in the VN Spec. I mean since we throw an error only if the namespace is same for both the Virtualservices sharing the same name. Believe we should be ok..

PR #332, opened 2020-08-05
Add integration tests and framework for virtualgateway and gatewayroute

fawadkhaliq:
*Description of changes:*
Add test framework changes and integration tests for virtualgateway and gatewayroute

### DIFF
--- a/test/framework/manifest/gatewayroute.go
+++ b/test/framework/manifest/gatewayroute.go
@@ -1,0 +1,71 @@
+package manifest
+
+import (
+	appmesh "github.com/aws/aws-app-mesh-controller-for-k8s/apis/appmesh/v1beta2"
+	"github.com/aws/aws-sdk-go/aws"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type GRBuilder struct {
+	Namespace string
+}
+
+func (b *GRBuilder) BuildGatewayRoute(instanceName string, vsName string, prefix string) *appmesh.GatewayRoute {
+	grName := b.buildServiceName(instanceName)
+
+	gr := &appmesh.GatewayRoute{
+		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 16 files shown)


test/framework/manifest/gatewayroute.go
```diff
@@ -0,0 +1,71 @@
...
+	appmesh "github.com/aws/aws-app-mesh-controller-for-k8s/apis/appmesh/v1beta2"
+	"github.com/aws/aws-sdk-go/aws"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type GRBuilder struct {
+	Namespace string
+}
+
+func (b *GRBuilder) BuildGatewayRoute(instanceName string, vsName string, prefix string) *appmesh.GatewayRoute {
```
achevuru, line 13: Looks like this function builds a GR with HTTP route by default. Can we name it as such and create an equivalent function for GRPC? Or pass in the protocol as a parameter to this function and then split off..

fawadkhaliq: +1, will adjust

test/framework/manifest/virtualgateway.go
```diff
@@ -0,0 +1,63 @@
...
+	return vg
+}
+
+func (b *VGBuilder) BuildVGListener(protocol appmesh.VirtualGatewayPortProtocol, port appmesh.PortNumber) appmesh.VirtualGatewayListener {
+	return appmesh.VirtualGatewayListener{
+		PortMapping: appmesh.VirtualGatewayPortMapping{
+			Port:     port,
+			Protocol: protocol,
+		},
+		HealthCheck: &appmesh.VirtualGatewayHealthCheckPolicy{
```
achevuru, line 45: Might help making these values configurable. Maybe we can have HealthCheck struct that the caller can fill in and pass it along to this function.

Issue #127, opened 2020-02-07
Allow modification of AWS Cloud Map DNS record TTL


quentinplessis:
When creating a virtual node with Cloud Map as the service discovery mechanism, the TTL of the DNS record registered [seems to be currently fixed to `300` seconds](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/1ea35cb834a0f42b43b53788b3b85d86f27c785f/pkg/aws/cloudmap.go#L86).

It would be quite useful to be able to change that value, either by giving to app-mesh-controller a unique value for all virtual nodes, or better by being able to specify it in the `VirtualNode` resource. 

```yaml
apiVersion: appmesh.k8s.aws/v1beta1
kind: VirtualNode
metadata:
  name: my-node
spec:
  meshName: my-mesh
  serviceDiscovery:
    cloudMap:
      namespaceName: my-namespace.local
      serviceName: my-service
      ttl: 10
```

To my understanding using Cloud Map is necessary when spanning App Mesh across several clusters. When deleting a pod, the pod IP address is deregistered from Cloud Map in a few seconds/dozen of seconds but requests keep being redirected to the now-nonexisting-pod's IP address for 5 minutes due to the TTL.

c-yokoyama:
I'm on the same issue, and hope that the issue will be resolved soon.

kiranmeduri:
@quentinplessis @c-yokoyama is this a global setting or you need it per virtual-node?

quentinplessis:
@kiranmeduri Thank you for your reply ! A global setting would be enough for my specific use case.

c-yokoyama:
In our case, a global setting would be enough, too.

achevuru:
#337

rbg:
while this has been fixed, it is not clear to me how i would set this value. If you could provide an example that would be appreciated.

PR #337, opened 2020-08-07
Support to configure DNS TTL value of CloudMap services

achevuru:
*Issue #, if available:*

Allow modification of AWS Cloud Map DNS record TTL #127

*Description of changes:*
 
-  Support to configure DNS TTL value of CloudMap services. "cloudmap-dns-ttl" value (in seconds) can be assigned/modified based on the requirement.
- If "cloudmap-dns-ttl" value is not provided to controller during bootstrap, a default value of 300s will be used.

### DIFF
--- a/pkg/cloudmap/resource_manager.go
+++ b/pkg/cloudmap/resource_manager.go
@@ -334,7 +334,7 @@ func (m *defaultResourceManager) createCloudMapServiceUnderPrivateDNSNamespace(c
 			DnsRecords: []*servicediscovery.DnsRecord{
 				{
 					Type: awssdk.String(servicediscovery.RecordTypeA),
-					TTL:  &cloudMapTTL,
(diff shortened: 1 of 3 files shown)


pkg/cloudmap/resource_manager.go
```diff
@@ -330,7 +334,7 @@ func (m *defaultResourceManager) createCloudMapServiceUnderPrivateDNSNamespace(c
 			DnsRecords: []*servicediscovery.DnsRecord{
 				{
 					Type: awssdk.String(servicediscovery.RecordTypeA),
-					TTL:  awssdk.Int64(defaultServiceDNSConfigTTL),
+					TTL:  &cloudMapTTL,
```
fawadkhaliq, line 337: Better to keep using the `awssdk.Int64` interface like you already had

Issue #328, opened 2020-08-04
Can not access sts in a private eks cluster


SeliMeli:
When using appmesh at private eks (cluster dont have access to internet), at the set up of appmesh controller, it will try to get account id from sts.amazonaws.com which is not accessable. But we have created an endpiont at vpc with private dns (sts.region.amazonaws.com). Is there any way to change the hostname when appmesh controller set up?
![image](https://user-images.githubusercontent.com/9347182/89282773-99a25f00-d67e-11ea-8cd6-d618213f7500.png)
(and set aws-account-id not work nither at deploy yaml

fawadkhaliq:
@SeliMeli thanks for filing the issue. The account id issue has been fixed here #330 

For the STS endpoint, appmesh controller does not currently support overriding it. We will add the support for overriding the endpoint

SeliMeli:
I think the point is the controller ignored region config I passed to it. Because normally, when region is set, it should use region host instead of global host.

fawadkhaliq:
@SeliMeli @jane829 we will expose the STS endpoint in 1.1.1 release. However, you might run into limitations in terms of what App Mesh supports with VPC endpoints. Please see this issue [1] for details. And please upvote this [1] if you would like support for VPC endpoints for App Mesh

[1] https://github.com/aws/aws-app-mesh-roadmap/issues/208

fawadkhaliq:
You can override STS endpoint using https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/339. Please note App Mesh for private EKS clusters will not work unless this [1] is resolved. Please consider upvoting 

[1] aws/aws-app-mesh-roadmap#208

fawadkhaliq:
The original issue #330 in the chain wasn't the reason to reach the global STS endpoint. #339  was added to provide a way to override the STS endpoint of choice, however, this can also be achieved using PrivateLink, where it will set the dns for public STS endpoint in the VPC. We will recommend going with the latter approach and rely on PrivateLink endpoints that should match the same DNS names as the public regional endpoints, therefore reverting this

PR #339, opened 2020-08-12
Add STS endpoint resolver to override the default

fawadkhaliq:
*Issue* #328 

*Description of changes:*
Add the support to override the default STS endpoint. 

Verified endpoint can be overridden, incorrect endpoint leads to controller throwing error, region mismatch leads to valid errors, when nothing is specified, default endpoints are used. In addition, all the subsequent API calls work fine for AppMesh and CloudMap.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -58,8 +58,6 @@ github.com/asaskevich/govalidator v0.0.0-20190424111038-f61b66f89f4a/go.mod h1:l
 github.com/asaskevich/govalidator v0.0.0-20200108200545-475eaeb16496 h1:zV3ejI06GQ59hwDQAvmK1qxOQGB3WuVTRoY0okPTAv0=
 github.com/asaskevich/govalidator v0.0.0-20200108200545-475eaeb16496/go.mod h1:oGkLhpf+kjZl6xBf758TQhh5XrAeiJv/7FRz/2spLIg=
 github.com/aws/aws-sdk-go v1.15.11/go.mod h1:mFuSZ37Z9YOHbQEwBWztmVzqXrEkub65tZoCYDt7FT0=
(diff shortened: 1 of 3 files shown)


M00nF1sh:
Hi Fawad,
Do we really need this?
`awsCfg = aws.NewConfig().WithRegion(cfg.Region).WithSTSRegionalEndpoint(endpoints.RegionalSTSEndpoint)` should be enough.
I think it don't work before because of the bug in https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/330

fawadkhaliq:
> Hi Fawad,
> I don't think we need this, override STS endpoint is hacky.
> 
> `awsCfg = aws.NewConfig().WithRegion(cfg.Region).WithSTSRegionalEndpoint(endpoints.RegionalSTSEndpoint)` should be enough.
> I think it don't work before might because of either below:
> 
> 1. the bug in #330, so the region passed in didn't take effect.
> 2. `sess = sess.Copy(awsCfg)` didn't copy the `WithSTSRegionalEndpoint(endpoints.RegionalSTSEndpoint)` settings, if so, we just need to set it again.

Yes, I thought about it. The original issue was because accountID wasn't passed (#330). This change was added more to support if customer is coming up with a "custom" endpoint (not the default PrivateLink override to the regional sts endpoint). I see your point. The sts endpoint used was global. If this was truly only happening because of region being overriden to account id, we can conside not adding the support to override it at all and have customers only rely on privateLink endpoints for regional sts (accessible from private clusters). Let me double check and then we can revert it or keep it otherwise

fawadkhaliq:
The original issue #330 wasn't the reason to reach the global STS endpoint. This change provides a way to override the STS endpoint of choice, however, this can also be achieved using PrivateLink, where it will set the dns for public STS endpoint in the VPC. We will recommend going with the latter approach and rely on PrivateLink endpoints that should match the same DNS names as the public regional endpoints

PR #343, opened 2020-08-16
CloudMap Integration tests

achevuru:
*Issue #, if available:*

*Description of changes:*
Support for CloudMap Integration tests
- Create and Delete of a CloudMap Namespace
- Create and Delete of a CloudMap Service
- Register and Deregister instances. Verify Instance attributes in CloudMap vs Local instances.

### DIFF
--- a/test/framework/resource/virtualnode/manager.go
+++ b/test/framework/resource/virtualnode/manager.go
@@ -2,6 +2,10 @@ package virtualnode
 
 import (
 	"context"
+	"fmt"
+	awssdk "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/servicediscovery"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"time"
 
 	appmesh "github.com/aws/aws-app-mesh-controller-for-k8s/apis/appmesh/v1beta2"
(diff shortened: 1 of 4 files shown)


test/integration/virtualnode/virtualnode_test.go
```diff
@@ -149,6 +161,101 @@ var _ = Describe("VirtualNode", func() {
...
+
+			By(fmt.Sprintf("create a deployment for VirtualNode"), func() {
+				dp := mb.BuildDeployment(vnName, 2, defaultAppImage, AppContainerPort, []corev1.EnvVar{})
+				err := f.K8sClient.Create(ctx, dp)
+				Expect(err).NotTo(HaveOccurred())
+				vnTest.Deployments[vnName]=dp
+			})
+
+			//Let Instances sync with CloudMap and Pod Readiness Gate go through
+			time.Sleep(30*time.Second)
```
fawadkhaliq, line 252: There's no better way to check by check-wait-retry-timeout instead of just sleep here? : (
This could end up being a flaky test when we enable it in the pipeline

test/integration/virtualnode/virtualnode_test.go
```diff
@@ -149,6 +161,101 @@ var _ = Describe("VirtualNode", func() {
 
 		})
 
+		It("should create a virtual node in CloudMap", func() {
```
fawadkhaliq, line 164: nit: should create a virtual node with CloudMap discovery

test/framework/resource/virtualnode/manager.go
```diff
@@ -25,20 +29,41 @@ type Manager interface {
...
 	}
 }
 
 type defaultManager struct {
 	k8sClient  client.Client
 	appMeshSDK services.AppMesh
+	cloudMapSDK services.CloudMap
 }
 
+const (
```
fawadkhaliq, line 49: Would be nice not to duplicate and use the constants defined in one place from source package

test/integration/virtualnode/virtualnode_test.go
achevuru: I mean that is what I initially started with but it was taking at least 15s+ and for that we had to introduce a delay before we reach out to cloudMap again. So, we had to introduce a sleep in some fashion inside the retry loop and the idea was why loop if we are sleeping anyways inside the retry loop and instead go with this.

test/integration/virtualnode/virtualnode_test.go
achevuru: Done

test/framework/resource/virtualnode/manager.go
achevuru: Sure, makes sense.

Issue #353, opened 2020-10-01
Jaeger tracing broken with App Mesh Envoy v1.15.0.0-prod


Y0Username:
While using App Mesh Envoy version `v1.15.0.0-prod` Jaeger tracing is broken due to regression.

Envoy is looking for the Jaeger tracing config in the env variable `ENVOY_TRACING_CFG_FILE`.
But the k8s controller puts the config in the env variable pointed by `ENVOY_STATS_CONFIG_FILE`.

The relevant code is [here](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/fa911c740ffcf835a44a4ef71346b152308ffc41/pkg/inject/envoy.go#L54)

The fix is to to pass the Jaeger tracing config for Envoy version `v1.15.0.0-prod` using the env variable `ENVOY_TRACING_CFG_FILE`

fawadkhaliq:
App Mesh Envoy bootstrap in Envoy version `v1.12.5.0-prod` was merging tracing config from `ENVOY_STATS_CONFIG_FILE` and this was changed in Envoy `v1.15.0.0-prod`. 

Controller will be updated to use the flag `ENVOY_TRACING_CFG_FILE` in 1.2.0. Workaround is to use Envoy version `v1.12.5.0-prod`

achevuru:
#359

PR #359, opened 2020-10-08
Fix for Jaeger tracing with Envoy v1.15.0

achevuru:
*Issue #, if available:*
#353

*Description of changes:*

Envoy v1.15.0 is expecting the Jaeger tracing config file to be specified under the env variable 'ENVOY_TRACING_CFG_FILE' while K8S controller currently specifies the config through the env variable 'ENVOY_STATS_CONFIG_FILE'. This commit addresses it along with some changes required for the Jaeger config file.

### DIFF
--- a/pkg/inject/jaeger.go
+++ b/pkg/inject/jaeger.go
@@ -13,13 +13,14 @@ tracing:
    "@type": type.googleapis.com/envoy.config.trace.v2.ZipkinConfig
    collector_cluster: jaeger
    collector_endpoint: "/api/v1/spans"
+   collector_endpoint_version: HTTP_JSON
    shared_span_context: false
 static_resources:
   clusters:
   - name: jaeger
     connect_timeout: 1s
-    type: strict_dns
-    lb_policy: round_robin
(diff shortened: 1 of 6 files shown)


pkg/inject/jaeger.go
```diff
@@ -13,13 +13,14 @@ tracing:
    "@type": type.googleapis.com/envoy.config.trace.v2.ZipkinConfig
    collector_cluster: jaeger
    collector_endpoint: "/api/v1/spans"
+   collector_endpoint_version: HTTP_JSON
    shared_span_context: false
 static_resources:
   clusters:
   - name: jaeger
     connect_timeout: 1s
-    type: strict_dns
```
fawadkhaliq, line 21: Looks good. You will probably need the same change for datadog tracing config[1]

[1] https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/inject/datadog.go#L21

fawadkhaliq (approved):
Thanks for updating the datadog config as well. Looks good!

Issue #361, opened 2020-10-08
ServiceDiscovery should be optional for VirtualNodes when listeners are not specified


achevuru:
ServiceDiscovery should be optional for VirtualNodes when listeners are not specified

PR #362, opened 2020-10-08
Fix VirtualNode's ServiceDiscovery Validation Logic

achevuru:
*Issue #, if available:*
#361

*Description of changes:*

Corrects an issue with recently introduced VirtualNode validation logic to make ServiceDiscovery optional when listeners are not specified. Also, included an additional check in the VirtualNode Validation flow to make sure ServiceDiscovery is specified when a listener is present in the VirtualNode spec.

### DIFF
--- a/webhooks/appmesh/virtualnode_validator.go
+++ b/webhooks/appmesh/virtualnode_validator.go
@@ -31,6 +31,9 @@ func (v *virtualNodeValidator) Prototype(req admission.Request) (runtime.Object,
 
 func (v *virtualNodeValidator) ValidateCreate(ctx context.Context, obj runtime.Object) error {
 	vn := obj.(*appmesh.VirtualNode)
+	if err := v.checkForRequiredFields(vn); err != nil {
+		return err
+	}
 	if err := v.checkVirtualNodeBackendsForDuplicates(vn); err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


webhooks/appmesh/virtualnode_validator.go
```diff
@@ -95,6 +102,14 @@ func (v *virtualNodeValidator) checkVirtualNodeBackendsForDuplicates(vn *appmesh
 	return nil
 }
 
+func (v *virtualNodeValidator) checkForRequiredFields(vn *appmesh.VirtualNode) error {
```
fawadkhaliq, line 105: Looks good. Let's also update the API docs [1] to align with the expected behavior [2] that if there's no inbound traffic, service discovery is optional. Updating the docs will be reflected here [3] when we release so customers are aware of the behavior


[1] https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/apis/appmesh/v1beta2/virtualnode_types.go#L335
[2] https://docs.aws.amazon.com/app-mesh/latest/userguide/virtual_nodes.html
[3] https://aws.github.io/aws-app-mesh-controller-for-k8s/reference/api_spec/#appmesh.k8s.aws/v1beta2.VirtualNodeSpec

webhooks/appmesh/virtualnode_validator.go
```diff
@@ -62,7 +68,8 @@ func (v *virtualNodeValidator) enforceFieldsImmutability(vn *appmesh.VirtualNode
 	if !reflect.DeepEqual(vn.Spec.MeshRef, oldVN.Spec.MeshRef) {
 		changedImmutableFields = append(changedImmutableFields, "spec.meshRef")
 	}
-	if oldVN.Spec.ServiceDiscovery.AWSCloudMap != nil && !reflect.DeepEqual(vn.Spec.ServiceDiscovery.AWSCloudMap, oldVN.Spec.ServiceDiscovery.AWSCloudMap) {
+	if oldVN.Spec.ServiceDiscovery != nil && oldVN.Spec.ServiceDiscovery.AWSCloudMap != nil &&
```
fawadkhaliq, line 71: Let's please extend the virtual node integration to test for cases where service discovery is not specified

fawadkhaliq (changes requested):

fawadkhaliq (approved):
/lgtm

Issue #352, opened 2020-09-29
Feature Request: Support STATSD-ADDRESS and other envoy configs support 


shwetasahuit:
The current [app-mesh-controller configuration](https://github.com/aws/eks-charts/blob/master/stable/appmesh-controller/README.md#configuration) doesn't contain all the  [envoy configs supported by App Mesh](https://docs.aws.amazon.com/app-mesh/latest/userguide/envoy.html#envoy-config). Update the App Mesh Controller to provide support for these envoy configs.

PR #365, opened 2020-10-12
Envoy configurability in AppMesh Controller

achevuru:
*Issue #, if available:*
#352

*Description of changes:*
  Extends Envoy configurability in the controller to support all the options that are supported in upstream AWS AppMesh. This check-in extends support for the below options:
- ENVOY_ADMIN_ACCESS_PORT
- ENVOY_ADMIN_ACCESS_LOG_FILE
- XRAY_DAEMON_PORT
- DATADOG_TRACER_PORT
- DATADOG_TRACER_ADDRESS
- STATSD_PORT
- STATSD_ADDRESS


pkg/inject/virtualgateway_envoy.go
```diff
@@ -15,6 +15,8 @@ const envoyVirtualGatewayEnvMap = `
   "APPMESH_VIRTUAL_NODE_NAME": "mesh/{{ .MeshName }}/virtualGateway/{{ .VirtualGatewayName }}",
   "APPMESH_PREVIEW": "{{ .Preview }}",
   "ENVOY_LOG_LEVEL": "{{ .LogLevel }}",
+  "ENVOY_ADMIN_ACCESS_PORT": "{{ .AdminAccessPort }}",
+  "ENVOY_ADMIN_ACCESS_LOG_FILE": "{{ .AdminAccessLogFile }}",
```
fawadkhaliq, line 19: Need `XrayDaemonPort` here as well?

pkg/inject/envoy.go
```diff
@@ -61,6 +69,22 @@ const envoyContainerTemplate = `
     {
       "name": "ENABLE_ENVOY_XRAY_TRACING",
       "value": "1"
+    }{{ end }}{{ if .EnableXrayTracing }},
+    {
+      "name": "XRAY_DAEMON_PORT",
+      "value": "{{ .XrayDaemonPort }}"
```
fawadkhaliq, line 75: Since we configure the xray sidecar and inject it, this needs to be changed in xray injector as well [1]

[1] https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/inject/xray.go#L19

pkg/inject/envoy.go
```diff
@@ -61,6 +69,22 @@ const envoyContainerTemplate = `
...
+    }{{ end }}{{ if .EnableDatadogTracing }},
+    {
+      "name": "ENABLE_ENVOY_DATADOG_TRACING",
+      "value": "1"
+    }{{ end }}{{ if .EnableDatadogTracing}},
+    {
+      "name": "DATADOG_TRACER_PORT",
+      "value": "{{ .DatadogTracerPort }}"
```
fawadkhaliq, line 83: We already configure the datadog address and port here [1], are these going to override the static resource config?


[1] https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/inject/datadog.go#L29

pkg/inject/envoy.go
```diff
@@ -61,6 +69,22 @@ const envoyContainerTemplate = `
     {
       "name": "ENABLE_ENVOY_XRAY_TRACING",
       "value": "1"
+    }{{ end }}{{ if .EnableXrayTracing }},
+    {
+      "name": "XRAY_DAEMON_PORT",
+      "value": "{{ .XrayDaemonPort }}"
+    }{{ end }}{{ if .EnableDatadogTracing }},
+    {
+      "name": "ENABLE_ENVOY_DATADOG_TRACING",
```
fawadkhaliq, line 78: Is this supposed to replace the static file based config we do here? [1]

[1] https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/inject/datadog.go

fawadkhaliq (changes requested):

fawadkhaliq:
containerPod for envoy also needs to be handled here [1] and at any other places where we used it

https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/inject/envoy.go#L24

pkg/inject/envoy.go
achevuru: More like other way around but we should instead make the above config optional.

pkg/inject/virtualgateway_envoy.go
achevuru: Yes. Maybe we should include all the tracing options (jaeger/datadog) in VG as well...

fawadkhaliq: +1, consider adding those in a separate PR as that would change the scope of this change

achevuru: Included XrayDaemonPort. Will add Jaeger and Datadog support via a separate PR.

pkg/inject/envoy.go
achevuru: After trying out both the modes, I believe we should stick to just passing the port and datadog tracer address. Default config that is being configured is same as what we send through the yaml. Also, this will give us the ability to configure the address and port instead of hardcoded values in the config file.

fawadkhaliq (approved):
Looks good. Thanks for addressing the comments. Please note that you'll need to update the `docs/guide/tracing.md` once you expose these flags via the Helm chart

Issue #347, opened 2020-09-13
Feature Request: Support for configurable connection pool settings in App Mesh


fawadkhaliq:
Configurable connection pool settings will help enable [circuit breaking](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/circuit_breaking) use-cases. AWS App Mesh will be adding [support for configurable connection pool settings](https://github.com/aws/aws-app-mesh-roadmap/issues/6). For integration with Kubernetes, add the support for connection pool settings in the App Mesh Kubernetes controller.

fawadkhaliq:
Connection pool support has been added in the controller. We'll be adding the support in Helm charts and examples as well.

fawadkhaliq:
Controller changes: https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/363, https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/366
Helm changes: https://github.com/aws/eks-charts/pull/295
Walkthrough: https://github.com/aws/aws-app-mesh-examples/pull/355

Issue #346, opened 2020-09-13
Feature Request: Envoy outlier detection support in App Mesh


fawadkhaliq:
[Outlier detection](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/outlier) is the process of determining whether some number of hosts in an upstream cluster are performing unlike the others and removing them from the healthy hosts. AWS App Mesh will be [adding support for Envoy outlier detection](https://github.com/aws/aws-app-mesh-roadmap/issues/6). For integration with Kubernetes, add the support for outlier detection in the App Mesh Kubernetes controller.

fawadkhaliq:
Outlier detection support has been added in the controller. We'll be adding the support in Helm charts and examples as well.

fawadkhaliq:
Controller changes: https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/356, https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/366
Helm changes: https://github.com/aws/eks-charts/pull/295
Walkthrough: https://github.com/aws/aws-app-mesh-examples/pull/354

PR #366, opened 2020-10-16
Add integration tests for outlier detection and connection pools

fawadkhaliq:
**Issue** #347 #346 

**Description of changes**
Add integration tests for outlier detection and connection pools
- Added outlier detection integration scenarios for virtual node
- Added connection pool integration scenarios for virtual node and virtual gateway
- Added validation webhook and unit test for virtual node and virtual gateway connection pools on create/update
- Added outlier detection and connection pool listener spec builder

_Note: tested with App Mesh [preview channel](https://docs.aws.amazon.com/app-mesh/latest/userguide/preview.html) as outlier detection and connection pools are currently only available in preview_


test/integration/virtualgateway/virtualgateway_test.go
```diff
@@ -361,5 +361,168 @@ var _ = Describe("VirtualGateway", func() {
 			})
 
 		})
+
+		It("Virtual Gateway Connection Pool Scenarios", func() {
```
achevuru, line 365: We might need feature tests for outlier detection and connection pools. It's definitely helpful to verify the API part but without feature tests we might've to manually test both the features.

Rest looks straightforward and good to me.

achevuru (approved):

test/integration/virtualgateway/virtualgateway_test.go
fawadkhaliq: Yes, absolutely - goes without saying. The additional traffic check scenarios are planned for e2e tests as follow ups

Issue #360, opened 2020-10-08
Deleting resources fails if controller malfunctioning


nomeelnoj:
If the controller is malfunctioning, or not yet up, and an app mesh resource is created (virtualservice, virtualrouter, etc.), they fail to create properly (as expected).

However, you cannot delete these resources once they are there--the delete command fails.  If you `kubectl get` the resource, there is no ARN listed (makes sense).

I have also tried deleting the entire namespace, but that hangs as well.  Seems to be an issue with the finalizers:
```
message: 'Some content in the namespace has finalizers remaining: finalizers.appmesh.k8s.aws/aws-appmesh-resources
      in 4 resource instances, finalizers.appmesh.k8s.aws/aws-cloudmap-resources in
      2 resource instances'
```

I am not sure what is going on here, but I am unable to delete resources that "failed" to create, and cannot update them, either.

Once the controller comes up (after the resources due to gitops flow) or because there was an issue with the controller, the logs show some interesting errors:

```
[...]

PR #367, opened 2020-10-19
Handle resource delete scenarios 

achevuru:
*Issue #, if available:*
#360
*Description of changes:*

Current controller implementation might prevent a user from deleting AppMesh resources if the corresponding AWS resource(AppMesh or CloudMap) creation failed for any reason (for ex: lack of required permissions). This PR aims to address that via using the resource ARN as a way to identify if the resource was created successfully to begin with. Controller updates the resource specs with their ARN upon a successful create. If a resource delete fails for any reason (other than NotFoundException) we check if the resource spec is populated with ARN, if not we assume the AWS resource creation never went through and proceed with the resource deletion.

Note: If updating ARN in the resource spec fails for any reason, this approach might result in a resource leak. However, this approach addresses most of the scenarios that might otherwise prevent an user from deleting unwanted resources (when a controller fails to delete the resource for any reason).

PR also fixes a minor issue with e2e TLS test for CloudMap.


pkg/gatewayroute/resource_manager.go
```diff
@@ -110,7 +110,10 @@ func (m *defaultResourceManager) Cleanup(ctx context.Context, gr *appmesh.Gatewa
 		return err
 	}
 	sdkGR, err := m.findSDKGatewayRoute(ctx, ms, vg, gr)
-	if sdkGR == nil {
+	if err != nil {
+		if sdkGR == nil || gr.Status.GatewayRouteARN == nil {
```
fawadkhaliq, line 114: `sdkGR == nil` check is also applicable for case when `err == nil`, I'd suggest checking this outside the `err != nil` statement so it covers for nil `sdkGR` regardless of error

Same applies to virtual gateway `sdkVG `

fawadkhaliq (changes requested):
One minor comment. Otherwise it looks good!

fawadkhaliq (approved):
/lgtm

PR #368, opened 2020-10-20
Add IAM policies for the controller and envoy

fawadkhaliq:
**Description of changes**
Added IAM policies for the controller and envoy proxies

Instead of using the managed policies `AWSAppMeshFullAccess`, `AWSCloudMapFullAccess`, this change adds the IAM permissions for the controller. Also, previously, we using the same IAM policies for the controller and Envoy proxies. This change also separates the Envoy IAM policy from the controller IAM policy. Helm charts will refer to these IAM policies in the installation steps.

**Testing done**
- Integration tests
- e2e tests
- App Mesh on EKS + Fargate

### DIFF
--- a/config/iam/controller-iam-policy.json
+++ b/config/iam/controller-iam-policy.json
@@ -1,0 +1,93 @@
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
+            "Action": [
+		"appmesh:ListVirtualRouters",
+		"appmesh:ListVirtualServices",
+		"appmesh:ListRoutes",
+		"appmesh:ListGatewayRoutes",
+		"appmesh:ListMeshes",
+		"appmesh:ListVirtualNodes",
+		"appmesh:ListVirtualGateways",
+		"appmesh:DescribeMesh",
+		"appmesh:DescribeVirtualRouter",
+		"appmesh:DescribeRoute",
+		"appmesh:DescribeVirtualNode",
+		"appmesh:DescribeVirtualGateway",
+		"appmesh:DescribeGatewayRoute",
+		"appmesh:DescribeVirtualService",
+		"appmesh:CreateMesh",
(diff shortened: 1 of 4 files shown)


config/iam/controller-iam-policy.json
```diff
@@ -0,0 +1,93 @@
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
+            "Action": [
```
bcelenza, line 6: A couple thoughts:

1. Can we use managed policies for this App Mesh instead of explicit? This will be easier on both customers and us when we make future changes. It's also safe to simplify this to: 

```
appmesh:Create*
appmesh:Describe*
appmesh:Update*
appmesh:Delete*
appmesh:List*
```

2. For the preview environment these will need to be `appmesh-preview`. We want a separate policy for that in general, since we experiment with things in preview (and wouldn't want to auto add to the prod policy). There are managed policies for preview as well.

config/iam/envoy-iam-policy.json
```diff
@@ -0,0 +1,20 @@
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
+            "Action": [
+                "appmesh:StreamAggregatedResources"
```
bcelenza, line 7: Similar to above, this will need a preview variant (`appmesh-preview:StreamAggregatedResources`).

The managed policies for Envoy *only* include App Mesh actions (no ACM for security reasons), so I don't think the managed policy approach really helps here.

config/iam/controller-iam-policy.json
fawadkhaliq: Thanks for the review. The reason for not adding wildcards in the `appmesh:Create*` etc is because it would add permissions (e.g CreateTags) or any other operations that might be added to AppMesh but the controller.

+1, I will add the preview variant separately

bcelenza (approved):

Issue #334, opened 2020-08-06
Introduce installation/startup checks in the controller


dastbe:
Looking at issues like #328, it would be nice if we could codify our expectations for running the controller (STS availability, App Mesh API availability, iptables modification, etc.) into a command that runs on controller installation or on-demand by customers to determine the suitability of their cluster. Not only does it give us the opportunity to give customers early warning about whether their installation will work, but also gives us a mechanism to proactively direct them to remediation steps.

achevuru:
Controller does have installation time validation builtin. Current behavior/validation is summarized below.

- Validates all the configurable parameters passed to the controller
- Validates multiple tracers are not enabled at the same time ( https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/inject/config.go#L76)
- Errors out if incorrect API version is used in App Mesh CRD/Specs
- Will exit if an incorrect STS endpoint is specified (indirectly validates values passed in for Region and Account ID)
- Logs valid error messages in controller logs, if required permissions are missing

Controller is not involved with any iptable related operations and therefore doesn’t require access to modify iptables. Controller is also not running across all the worker nodes in a cluster. ‘Proxy Init’ container that gets included as an init container in each of the pod specs will be assigned appropriate permissions for it to be able to setup required iptable rules for envoy.

PR #374, opened 2020-11-01
Installation checks for controller

achevuru:
*Issue #, if available:*
#334
*Description of changes:*

  Controller will now validate if it has relevant permissions to access AWS App Mesh and CloudMap services during bootup. If permissions to access App Mesh are missing, controller will print an error message and will exit. If CloudMap permissions are missing, controller will print an error message and will move on as CloudMap access is not mandatory. Although this validation is turned on by default, one can disable this via a configurable parameter (enable-aws-access-validation).

### DIFF
--- a/main.go
+++ b/main.go
@@ -69,6 +69,7 @@ func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
 	var enableCustomHealthCheck bool
+	var enableAWSAccessValidation bool
 	var logLevel string
 	awsCloudConfig := aws.CloudConfig{ThrottleConfig: throttle.NewDefaultServiceOperationsThrottleConfig()}
 	injectConfig := inject.Config{}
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -124,6 +127,17 @@ func main() {
 		os.Exit(1)
 	}
 
+	if enableAWSAccessValidation {
+		appMeshAccessError, cloudMapAccessError := cloud.Validate()
+		if cloudMapAccessError != nil {
+			setupLog.Error(err, "CloudMap access failed. Required permissions missing. Refer: https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/config/iam/")
```
fawadkhaliq, line 133: Since CloudMap is optional, consider changing this to warning instead of error

pkg/aws/cloud.go
```diff
@@ -91,3 +97,29 @@ func (c *defaultCloud) AccountID() string {
 func (c *defaultCloud) Region() string {
 	return c.cfg.Region
 }
+
+func (c *defaultCloud) Validate() (appMeshAccessError, cloudMapAccessError error) {
```
fawadkhaliq, line 101: I came across SimulatePrincipalPolicy API [1] that allows programatic way to validate and audit permissions for IAM users, groups, and roles. Should we use `SimulatePrincipalPolicy` instead with the list of `ActionNames` and checking their `EvalDecision`? 

[1] https://docs.aws.amazon.com/sdk-for-go/api/service/iam/#IAM.SimulatePrincipalPolicy

Issue #395, opened 2020-11-30
Sidecar Inject/Replacement fails when x-ray tracing is enabled


marceloboeira:
**Describe the bug**

When trying to use https://github.com/aws/aws-app-mesh-examples/blob/master/walkthroughs/howto-k8s-connection-pools/v1beta2/manifest.yaml.template#L274 with `tracing.enabled=true` && `tracing.provider=x-ray` the pod fails.

**Steps to reproduce**

1. Follow https://github.com/aws/aws-app-mesh-examples/blob/master/walkthroughs/howto-k8s-connection-pools/
1. Enable x-ray tracing for appmeh-controller
1. Restart deployments to pick-up changes
1. See that any deployment with `controller-override` will fail to replace pods and error with a missing `containerPort` definition.

**Expected outcome**

Just work as before...

**Environment**

* App Mesh controller version: 1.2
* Envoy version
* Are you using any integrations? X-ray, Jaeger etc. If so versions? Yes, latest x-ray
* Kubernetes version: 1.18
* Using EKS (yes/no), if so version? 4

**Additional Context**:

fawadkhaliq:
@marceloboeira thanks for reporting the issue.

The issue was introduced here: https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/365. xray daemon port is not configured when injecting xray for virtual gateway pods: https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/inject/inject.go#L169

*Impact*
This impact virtual gateway pod specs marked for xRay sidecar injection. There's no impact on virtual nodes

*Workaround*
Specify the xRay sidecar in the virtual gateway pod spec and not rely on the mutator. You could do this by adding the following container to the virtual gateway pod spec. xRay injector does not override the existing xRay container for users who want to bring their our xRay sidecars. 

```
name: xray-daemon
image: amazon/aws-xray-daemon
securityContext:
  runAsUser: 1337
ports:
- containerPort: 2000
  name: xray
  protocol: UDP
env:
- name: AWS_REGION
  value: us-west-2
```

We'll have a fix shortly

marceloboeira:
@fawadkhaliq I was diving into the code yesterday trying to find out where was the issue, and yes, by manually placing the container it worked. Thanks for the fix!

fawadkhaliq:
> One last question is, the `tracing.port` set in the helm chart is ignored when using xray right?

[...]

PR #396, opened 2020-12-01
Add xray daemon port in virtualgateway pod mutator

fawadkhaliq:
**Issue**
#395 

**Description of changes**
When xray sidecar injection (tracing) is enabled, virtual gateway pod mutator does not pass the xray daemon port to the mutator. This leads to xray daemon container coming up without a container port. The issue was introduced as part of adding new config parameters for Envoy (https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/365) where xray mutotar init for virtual gateway does not get the xray daemon port passed down to the mutator, which leads to incorrect spec. 

This change adds the `XrayDaemonPort` to the xray mutator init, adds checks for mandatory fields and corresponding tests to avoid regression


marceloboeira (approved):
🎉

achevuru (approved):

PR #404, opened 2020-12-16
[Preview]mTLS-SDS based e2e integration tests

achevuru:
*Issue #, if available:*
NA
*Description of changes:*

PR introduces e2e integration test support for SDS based mTLS feature. Right now, the test is tailored for SPIRE as the SDS provider but should be easy to modify it to support other providers.

### DIFF
--- a/test/e2e/fishapp/certs/spire_setup.yaml
+++ b/test/e2e/fishapp/certs/spire_setup.yaml
@@ -1,0 +1,307 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: spire
+---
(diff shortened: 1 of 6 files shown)


test/e2e/fishapp/certs/spire_setup.yaml
```diff
@@ -0,0 +1,307 @@
+apiVersion: v1
```
fawadkhaliq, line 1: Once we introduce a helm chart for SPIRE, it'd be good to integrate that into the tests instead

fawadkhaliq (approved):

PR #418, opened 2021-01-13
Support to control/toggle SDS  per VirtualNode

achevuru:
*Issue #, if available:*
NA
*Description of changes:*

Ability to enable SDS based mTLS is currently a controller level flag. This PR, adds support to disable SDS @ VirtualNode level via an annotation on the corresponding deployment spec. AppMesh, currently bootstraps a static SDS cluster purely based on the presence of an environment variable (and not dynamically based on the presence of mTLS SDS config) and this change will help avoid it at the controller level.

If SDS is enabled via the controller flag(and SDS UDS path specified), controller will populate the "APPMESH_SDS_SOCKET_PATH" env variable and mount the UDS in to Envoy. To disable this @ VirtualNode level, one can use annotation "appmesh.k8s.aws/sds" set to disabled on the corresponding deployment spec.

### DIFF
--- a/pkg/inject/utils.go
+++ b/pkg/inject/utils.go
@@ -8,6 +8,10 @@ import (
 	"text/template"
 )
 
+const (
+	AppMeshSDSSocketVolume = "appmesh-sds-socket-volume"
+)
+
 func renderTemplate(name string, t string, meta interface{}) (string, error) {
 	tmpl, err := template.New(name).Parse(t)
 	if err != nil {
@@ -90,10 +94,22 @@ func isSDSDisabled(pod *corev1.Pod) bool {
 	return false
 }
 
+func isSDSVolumePresent(pod *corev1.Pod, SdsUdsPath string) bool {
+	for _, volume := range pod.Spec.Volumes {
+		if volume.HostPath != nil && volume.HostPath.Path == SdsUdsPath {
+			return true
+		}
+	}
+	return false
+}
+
 func mutateSDSMounts(pod *corev1.Pod, envoyContainer *corev1.Container, SdsUdsPath string) {
 	SDSVolumeType := corev1.HostPathSocket
+	if isSDSVolumePresent(pod, SdsUdsPath) {
+		return
+	}
 	volume := corev1.Volume{
-		Name: "sds-socket-volume",
+		Name: AppMeshSDSSocketVolume,
 		VolumeSource: corev1.VolumeSource{
 			HostPath: &corev1.HostPathVolumeSource{
 				Path: SdsUdsPath,
@@ -103,7 +119,7 @@ func mutateSDSMounts(pod *corev1.Pod, envoyContainer *corev1.Container, SdsUdsPa
 	}
 
 	volumeMount := corev1.VolumeMount{
-		Name:      "sds-socket-volume",
+		Name:      AppMeshSDSSocketVolume,
 		MountPath: SdsUdsPath,
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/inject/utils.go
```diff
@@ -80,3 +80,33 @@ func containsEnvoyContainer(pod *corev1.Pod) (bool, int) {
 	}
 	return false, -1
 }
+
+func isSDSDisabled(pod *corev1.Pod) bool {
+	if v, ok := pod.ObjectMeta.Annotations[AppMeshSDSAnnotation]; ok {
+		if v == "disabled" {
```
fawadkhaliq, line 86: Let's log a warning when user configures an unsupported value like "enabled"

pkg/inject/utils.go
```diff
@@ -80,3 +80,33 @@ func containsEnvoyContainer(pod *corev1.Pod) (bool, int) {
...
+			return true
+		}
+	}
+	return false
+}
+
+func mutateSDSMounts(pod *corev1.Pod, envoyContainer *corev1.Container, SdsUdsPath string) {
+	SDSVolumeType := corev1.HostPathSocket
+	volume := corev1.Volume{
+		Name: "sds-socket-volume",
```
fawadkhaliq, line 96: nit: prefix the name with "appmesh" so the name is not super generic and can be identified for debugging etc

pkg/inject/utils.go
```diff
@@ -80,3 +80,33 @@ func containsEnvoyContainer(pod *corev1.Pod) (bool, int) {
...
+		},
+	}
+
+	volumeMount := corev1.VolumeMount{
+		Name:      "sds-socket-volume",
+		MountPath: SdsUdsPath,
+	}
+
+	envoyContainer.VolumeMounts = append(envoyContainer.VolumeMounts, volumeMount)
+	pod.Spec.Volumes = append(pod.Spec.Volumes, volume)
```
fawadkhaliq, line 111: To make sure this is idempotent, if a volume is already there, we should not append it again

pkg/inject/virtualgateway_envoy.go
```diff
@@ -128,6 +128,9 @@ func (m *virtualGatewayEnvoyConfig) buildTemplateVariables(pod *corev1.Pod) Virt
 	meshName := m.getAugmentedMeshName()
 	virtualGatewayName := aws.StringValue(m.vg.Spec.AWSName)
 	preview := m.getPreview(pod)
+	if m.mutatorConfig.enableSDS && isSDSDisabled(pod) {
+		m.mutatorConfig.enableSDS = false
```
fawadkhaliq, line 132: This reads a bit off changing state of `m.mutatorConfig.enableSDS` for the entire package. i.e. it leaks the bounds of this specific function. It is better to use separate function local variable that is passed as a value of `EnableSDS ` in `VirtualGatewayEnvoyVariables`

Same applies to the change in envoy.go

fawadkhaliq (changes requested):

pkg/inject/virtualgateway_envoy.go
achevuru: This setting is per Pod (i.e.,) inject config is populated with defaults per pod and we want to toggle the SDS support for Pod level anyways.

pkg/inject/utils.go
achevuru: Sure, will change it.

pkg/inject/utils.go
achevuru: Yeah, good point.

pkg/inject/utils.go
achevuru: Agree, for VirtualGateway scenario. We can do a name based match to make sure we don't already have an existing volume.

kishorj (approved):

Issue #371, opened 2020-10-27
[Enhancement] Optimize memory usage in AppMesh controller


achevuru:
AppMesh controller follows the standard List+Watch model to keep track of all the relevant resources (i.e.,) Mesh, VirtualNodes, VirtualServices, VirtualRouters and Pods. client-go lib will cache the entire object for each of these resources in the cache. 

**Why is this a problem?**

As we scale each of these resources in the cluster, memory usage goes up proportionately. Larger than normal Pod objects in general will exacerbate the problem. Controller currently sets the default memory limit of 1 Gig. You can change it based on your scale use

**Proposal**

 While the controller needs access to the entire AppMesh resource (VirtualNodes, VirtualServices, VirtualRouters, Mesh) specs, the same is not true for Pod Objects. Controller can be smart in that it can only cache required portions of Pod specs and strip out the details that it doesn't care about. This will help to bring down the memory usage and also will bring predictability around controller's memory usage as memory usage will no longer depend on the size of the Pod Object (i.e.,) memory usage per pod object will be same no matter the size of the Pod object.

PR #419, opened 2021-01-13
Optimize Memory Usage for AppMesh Controller

cgchinmay:
*Issue #371

*Description of changes:*
Added a custom_controller for optimizing Pod spec and add it to the referenced datastore
Added a pod_conversion class which implements the Converter Interface to strip down pod objects
Added custom_source which is a custom event source which can allow to watch over Create,Update and Delete Channels

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   build:
     name: build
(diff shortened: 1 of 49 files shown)


cgchinmay:
Will Recreate the PR after fixing Integration Tests

PR #422, opened 2021-01-16
Optimize Memory Usage

cgchinmay:
*Issue #419

*Description of changes:*
Added a custom_controller for optimizing Pod spec and add it to the referenced datastore
Added a pod_conversion class which implements the Converter Interface to strip down pod objects
Added custom_source which is a custom event source which can allow to watch over Create,Update and Delete Channels
Updated Pod List Calls to make use of the custom DataStore


M00nF1sh (commented):
we don't need multiple source..i think a single one is better, which ties source & eventHandler together

pkg/conversions/pod_conversion.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// ResourceType to watch and list
+func (c *PodConverter) ResourceType() runtime.Object {
+	return c.K8sResourceType
+}
+
+// StripDownPod removes all the extra details from pod that are not
+// required by the controller.
+func (c *PodConverter) stripDownPod(pod *v1.Pod) *v1.Pod {
```
fawadkhaliq, line 78: You may want to re-evaluate the fields of the new pod spec/fields. I don't think we need all of the fields here. This repo holds two sub-systems: controllers and injectors/webhooks. This change is relevant for the former

cgchinmay:
> we don't need multiple source..i think a single one is better, which ties source & eventHandler together

Updated

pkg/conversions/pod_conversion.go
cgchinmay: Removed unnecessary fields

controllers/appmesh/cloudmap_controller.go
```diff
@@ -36,16 +35,23 @@ type cloudMapReconciler struct {
 	finalizerManager            k8s.FinalizerManager
 	cloudMapResourceManager     cloudmap.ResourceManager
 	enqueueRequestsForPodEvents handler.EventHandler
+	notificationChannel         chan interface{}
```
M00nF1sh, line 38: it's better to have three type-safe channel than a single non-type safe channel. 
explicit is always better than implicit

M00nF1sh: the channel declaration should be `chan <- "the type"` on receiver side.

controllers/appmesh/cloudmap_controller.go
```diff
@@ -60,10 +66,11 @@ func (r *cloudMapReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 }
 
 func (r *cloudMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
+
 	return ctrl.NewControllerManagedBy(mgr).
 		Named("cloudMap").
 		For(&appmesh.VirtualNode{}).
-		Watches(&source.Kind{Type: &corev1.Pod{}}, r.enqueueRequestsForPodEvents).
+		Watches(&customsource.CustomChannel{Source: r.notificationChannel}, r.enqueueRequestsForPodEvents).
```
M00nF1sh, line 73: i think both `customsource` and `custom` package should be replaced with the existing `k8s` package.
so it's like `k8s.NotificationChannel` and `k8s.PodController`

main.go
```diff
@@ -114,7 +127,21 @@ func main() {
...
+	// Channel that will be notified for the create, Update, delete events on pod object
+	podEventNotificationChannel := make(chan interface{})
+
+	// Custom data store, it should not be accessed directly as the cache could be out of sync
+	// on startup. Must be accessed from the pod controller's data store instead
+	dataStore := cache.NewIndexer(k8s.NSKeyIndexer, k8s.NamespaceIndexer())
+
+	kubeConfig := ctrl.GetConfigOrDie()
+	// Set the API Server QPS and Burst
+	kubeConfig.QPS = float32(custom.DefaultAPIServerQPS)
```
M00nF1sh, line 139: let's define DefaultAPIServerQPS and Burst in main.go since it's not specific to custom resource

controllers/appmesh/custom/custom_controller.go
```diff
@@ -0,0 +1,244 @@
...
+
+// Controller Interface implemented by NewController
+type Controller interface {
+	// StartController starts the controller. Will block the calling routine
+	StartController(dataStore cache.Indexer, stopChanel chan struct{})
+	// GetDataStore returns the data store once it has synced with the API Server
+	GetDataStore() cache.Indexer
+}
+
+// NewController is an Informer which converts Pod Objects and notifies corresponding even handlers via Channels
```
M00nF1sh, line 59: even`t`

controllers/appmesh/custom/custom_controller.go
```diff
@@ -0,0 +1,244 @@
...
+	// PageLimit is the number of objects returned per page on a list operation
+	PageLimit int64
+	// Namespace to list/watch for
+	Namespace string
+	// Converter is the converter implementation that converts the k8s
+	// object before storing in the data store
+	Converter Converter
+	// ResyncPeriod how often to sync using list with the API Server
+	ResyncPeriod time.Duration
+	// RetryOnError weather item should be retried on error. Should remain false in usual use case
```
M00nF1sh, line 72: whether

controllers/appmesh/custom/custom_controller.go
```diff
@@ -0,0 +1,244 @@
...
+	// StartController starts the controller. Will block the calling routine
+	StartController(dataStore cache.Indexer, stopChanel chan struct{})
+	// GetDataStore returns the data store once it has synced with the API Server
+	GetDataStore() cache.Indexer
+}
+
+// NewController is an Informer which converts Pod Objects and notifies corresponding even handlers via Channels
+type NewController struct {
+	// ClientSet is the kubernetes client set
+	ClientSet *kubernetes.Clientset
```
M00nF1sh, line 62: let's use lowercase for private members like `clientSet`

controllers/appmesh/custom/custom_controller.go
```diff
@@ -0,0 +1,244 @@
...
+						c.notifyChannelOnUpdate(old, convertedObj)
+					} else {
+						if err := c.dataStore.Add(convertedObj); err != nil {
+							return err
+						}
+						c.notifyChannelOnCreate(convertedObj)
+					}
+
+					if err != nil {
+						return err
```
M00nF1sh, line 124: when will this branch be invoked?

main.go
```diff
@@ -213,6 +264,22 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Only start the controller when the leader election is won
+	mgr.Add(manager.RunnableFunc(func(stop <-chan struct{}) error {
+		setupLog.Info("starting custom controller")
+
+		stopChannel := make(chan struct{})
```
M00nF1sh, line 271: i don't think we need to do an another stopChannel here. we should directly use the stop channel passed from controller. please sync with @abhipth on why it's needed.
also, we shouldn't close the stopChannel inside `StartController`, a component shouldn't close channel passed to it.

controllers/appmesh/custom/custom_controller.go
```diff
@@ -0,0 +1,244 @@
...
+					}
+					c.notifyChannelOnDelete(convertedObj)
+				}
+			}
+			return nil
+		},
+	}
+
+	c.Log.Info("starting custom controller")
+	defer close(stopChanel)
```
M00nF1sh, line 139: always a bad style in golang to close channel on receiver side .

main.go
```diff
@@ -76,6 +85,7 @@ func main() {
 	var enableLeaderElection bool
 	var enableCustomHealthCheck bool
 	var logLevel string
+	var listPageLimit int8
```
M00nF1sh, line 88: why int8 given it's int64 when been used?

main.go
```diff
@@ -123,6 +150,21 @@ func main() {
 		LeaderElectionID:       "appmesh-controller-leader-election",
 		HealthProbeBindAddress: healthProbeBindAddress,
 	})
+
+	podController := &custom.NewController{
```
M00nF1sh, line 154: let's have constructor func for controller to be defined along with the controller struct

pkg/customsource/custom_source.go
```diff
@@ -0,0 +1,152 @@
...
+type Source interface {
+	// Start is internal and should be called only by the Controller to register an EventHandler with the Informer
+	// to enqueue reconcile.Requests.
+	Start(handler.EventHandler, workqueue.RateLimitingInterface, ...predicate.Predicate) error
+}
+
+var _ Source = &CustomChannel{}
+
+// type EventType struct {
+// 	event interface{}
```
M00nF1sh, line 29: let's clean unused code here

pkg/customsource/custom_source.go
```diff
@@ -0,0 +1,152 @@
...
+
+// CustomChannel monitors channels of type Create/Update/Delete
+type CustomChannel struct {
+	// once ensures the event distribution goroutine will be performed only once
+	once sync.Once
+
+	// stop is to end ongoing goroutine, and close the Create channel
+	stop <-chan struct{}
+
+	Source chan interface{}
```
M00nF1sh, line 40: let's be explicit to have add/del/update source channel

pkg/k8s/wrapper.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"github.com/aws/aws-app-mesh-controller-for-k8s/controllers/appmesh/custom"
+	v1 "k8s.io/api/core/v1"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// K8sWrapper represents an interface with all the common operations on pod objects
```
M00nF1sh, line 14: this feels more like a PodRepository instead of k8swrapper :D

pkg/k8s/pod_utils.go
```diff
@@ -20,6 +24,33 @@ func GetPodCondition(pod *corev1.Pod, conditionType corev1.PodConditionType) *co
...
+func NamespaceIndexer() cache.Indexers {
+	indexer := map[string]cache.IndexFunc{}
+	indexer[Namespace] = func(obj interface{}) (strings []string, err error) {
+		return []string{obj.(*corev1.Pod).Namespace}, nil
+	}
+	return indexer
+}
+
+// NodeNameIndexer returns indexer to index in the data store using node name
+func NodeNameIndexer() cache.Indexers {
```
M00nF1sh, line 37: is this NodeNameIndexer been used?

pkg/k8s/pod_utils.go
```diff
@@ -20,6 +24,33 @@ func GetPodCondition(pod *corev1.Pod, conditionType corev1.PodConditionType) *co
...
+	indexer := map[string]cache.IndexFunc{}
+	indexer[NodeNameSpec] = func(obj interface{}) (strings []string, err error) {
+		return []string{obj.(*corev1.Pod).Spec.NodeName}, nil
+	}
+	return indexer
+}
+
+// NSKeyIndexer is the key function to index the pod object using namespace/name
+func NSKeyIndexer(obj interface{}) (string, error) {
```
M00nF1sh, line 46: this function `k8s.NSKeyIndexer` doesn't hint that it's specific to a pod yet the implementation is.
Let's do either below
1. change the function name to have pod
2. move this into another file like object_utils.go, and use `meta.Accessor(obj)` to decouple the implementation from pod, which can work for any k8s object. (preferred approach)

same comment for NamespaceIndexer above.

controllers/appmesh/custom/custom_controller.go
cgchinmay: Its used in main.go while initializing the struct. So only the exposed variables are Uppercase

controllers/appmesh/custom/custom_controller.go
cgchinmay: Redundant

pkg/k8s/wrapper.go
cgchinmay: Tried to keep the naming convention in parity with VPC resource controller. I will update the name though

pkg/k8s/pod_utils.go
cgchinmay: will remove

controllers/appmesh/cloudmap_controller.go
```diff
@@ -36,16 +36,30 @@ type cloudMapReconciler struct {
 	finalizerManager            k8s.FinalizerManager
 	cloudMapResourceManager     cloudmap.ResourceManager
 	enqueueRequestsForPodEvents handler.EventHandler
+	//notificationChannel         chan interface{}
```
M00nF1sh, line 39: remember to remove this unused code

go.mod
```diff
@@ -8,13 +8,15 @@ require (
 	github.com/go-logr/logr v0.1.0
 	github.com/golang/mock v1.4.3
 	github.com/google/go-cmp v0.4.0
+	github.com/nxadm/tail v1.4.6 // indirect
```
M00nF1sh, line 11: let's revert this change if we don't rely on it.

main.go
```diff
@@ -123,6 +150,21 @@ func main() {
 		LeaderElectionID:       "appmesh-controller-leader-election",
 		HealthProbeBindAddress: healthProbeBindAddress,
 	})
+
+	podController := &k8s.PodController{
+		ClientSet:    clientSet,
+		PageLimit:    int64(listPageLimit),
+		Namespace:    metav1.NamespaceAll,
+		ResyncPeriod: syncPeriod,
+		Queue:        cache.NewDeltaFIFO(k8s.PodNSKeyIndexer, dataStore),
```
M00nF1sh, line 159: k8s.PodNSKeyIndexer can be replaced with `cache.MetaNamespaceKeyFunc

main.go
```diff
@@ -114,7 +128,20 @@ func main() {
 		"BuildDate", version.BuildDate,
 	)
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	podEventNotificationChan := make(chan k8s.PodEvent)
+
+	// Custom data store, it should not be accessed directly as the cache could be out of sync
+	// on startup. Must be accessed from the pod controller's data store instead
+	dataStore := cache.NewIndexer(k8s.PodNSKeyIndexer, k8s.PodNamespaceIndexer())
```
M00nF1sh, line 135: k8s.PodNSKeyIndexer can be replaced with `cache.MetaNamespaceKeyFunc

pkg/k8s/pod_utils.go
```diff
@@ -20,6 +24,24 @@ func GetPodCondition(pod *corev1.Pod, conditionType corev1.PodConditionType) *co
 	return nil
 }
 
+// PodNamespaceIndexer returns indexer to index Pods in datastore using namespace
+func PodNamespaceIndexer() cache.Indexers {
```
M00nF1sh, line 28: let's have a NamespaceIndexFunc which uses meta.Accessor(obj), and it's not pod specific

pkg/k8s/pod_wrapper.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// k8sWrapper is the wrapper object with the client
+type podsWrapper struct {
+	podController *PodController
+	cacheClient   client.Client
+}
+
+// NewPodsWrapper returns a new PodsWrapper
+func NewPodsWrapper(client client.Client, podController *PodController) PodsWrapper {
```
M00nF1sh, line 26: let's reply on the datastore directly instead of reply on podController

M00nF1sh: otherwise, the datastore should be initialized within podController

main.go
```diff
@@ -123,6 +150,21 @@ func main() {
 		LeaderElectionID:       "appmesh-controller-leader-election",
 		HealthProbeBindAddress: healthProbeBindAddress,
 	})
+
+	podController := &k8s.PodController{
```
M00nF1sh, line 154: let's have a constructor func like
NewPodController(xxxxx...)

main.go
```diff
@@ -123,6 +150,21 @@ func main() {
...
+		ClientSet:    clientSet,
+		PageLimit:    int64(listPageLimit),
+		Namespace:    metav1.NamespaceAll,
+		ResyncPeriod: syncPeriod,
+		Queue:        cache.NewDeltaFIFO(k8s.PodNSKeyIndexer, dataStore),
+		Converter: &conversions.PodConverter{
```
M00nF1sh, line 160: let's have a constructor func for PodConverter as well

main.go
```diff
@@ -213,6 +264,20 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Only start the controller when the leader election is won
+	mgr.Add(manager.RunnableFunc(func(stop <-chan struct{}) error {
```
M00nF1sh, line 268: nit: ideally we should startController here, and explicitly wait for cache sync before call mgr.start.  See https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/main.go#L154

The benefit is the controller's will not operator until the cache is synced.
In current code which relies on mgr, controllers might work without waiting for the pod cache, and only blocks on when it need pod information.

pkg/k8s/custom_controller.go
```diff
@@ -0,0 +1,232 @@
...
+	return &cache.ListWatch{ListFunc: listFunc, WatchFunc: watchFunc}
+}
+
+// notifyChannelOnCreate notifies the add event on the appropriate channel
+func (c *PodController) notifyChannelOnCreate(obj interface{}) error {
+	meta, err := apimeta.Accessor(obj)
+	if err != nil {
+		return err
+	}
+	c.PodEventNotificationChan <- PodEvent{
```
M00nF1sh, line 190: let's have a GenericEvent

pkg/k8s/custom_source.go
```diff
@@ -0,0 +1,181 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
+)
+
+const (
+	// defaultBufferSize is the default number of event notifications that can be buffered.
+	defaultBufferSize = 1024
+)
+
+// Source is a source of events
+type Source interface {
```
M00nF1sh, line 22: we don't need to have our own use Source interface

pkg/k8s/custom_source.go
```diff
@@ -0,0 +1,181 @@
...
+	EventType
+
+	// Meta is the ObjectMeta of the Kubernetes Type that was created
+	Meta metav1.Object
+
+	// Object is the object from the event
+	Object runtime.Object
+
+	// MetaOld is the ObjectMeta of the Kubernetes Type that was updated (before the update)
+	MetaOld metav1.Object
```
M00nF1sh, line 50: Meta      -> for the create/update's new oldObj
Object   -> for the create/update's new oldObj

OldMeta     -> for the delete or update's old oldObj
OldObject  -> for the delete or update's old oldObj



for example k8s uses below.

```
         // Object is the object from the incoming request.
	// +optional
	Object runtime.RawExtension `json:"object,omitempty" protobuf:"bytes,9,opt,name=object"`
	// OldObject is the existing object. Only populated for DELETE and UPDATE requests.
	// +optional
	OldObject runtime.RawExtension `json:"oldObject,omitempty" protobuf:"bytes,10,opt,name=oldObject"`
```

pkg/k8s/pod_utils.go
```diff
@@ -4,10 +4,14 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/cache"
 )
 
 const (
 	ConditionAWSCloudMapHealthy = "conditions.appmesh.k8s.aws/aws-cloudmap-healthy"
+	NodeNameSpec                = "nodeName"
+	Namespace                   = "namespace"
```
M00nF1sh, line 14: Namespace to be "NamespaceIndexKey" in object_utils.go

main.go
```diff
@@ -123,6 +145,19 @@ func main() {
 		LeaderElectionID:       "appmesh-controller-leader-election",
 		HealthProbeBindAddress: healthProbeBindAddress,
 	})
+
+	converter := conversions.NewPodConverter("pods", &corev1.Pod{})
```
M00nF1sh, line 149: seems "pods", &corev1.Pod{} is not needed since it's a podconverter

pkg/conversions/pod_conversion.go
```diff
@@ -0,0 +1,99 @@
...
+
+	v1 "k8s.io/api/core/v1"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// PodConverter implements the interface to convert k8s pod object to a stripped down
+// version of pod to save on memory utilized
+type podConverter struct {
+	K8sResource     string
```
M00nF1sh, line 27: i don't think this two variable is needed.

pkg/k8s/custom_controller.go
```diff
@@ -0,0 +1,250 @@
...
+			// from oldest to newest
+			for _, d := range obj.(cache.Deltas) {
+				// Strip down the pod object and keep only the required details
+				convertedObj, err := c.converter.ConvertObject(d.Object)
+				if err != nil {
+					return err
+				}
+				switch d.Type {
+				case cache.Sync, cache.Added, cache.Updated:
+					if old, exists, err := c.dataStore.Get(convertedObj); err == nil && exists {
```
M00nF1sh, line 120: what if error != nil?

pkg/k8s/object_utils.go
```diff
@@ -0,0 +1,26 @@
...
+	"k8s.io/client-go/tools/cache"
+)
+
+const (
+	//NamespaceIndexKey to index objects in datastore
+	NamespaceIndexKey = "namespace"
+)
+
+// NamespaceKeyIndexerFunc returns indexers to index objects in datastore using namespace
+func NamespaceKeyIndexerFunc() cache.Indexers {
```
M00nF1sh, line 16: this should be NamespaceKeyIndexFunc

pkg/k8s/custom_controller.go
```diff
@@ -0,0 +1,250 @@
...
+		clientSet:                clientSet,
+		pageLimit:                pageLimit,
+		namespace:                namesspace,
+		converter:                converter,
+		resyncPeriod:             resyncPeriod,
+		retryOnError:             retryOnError,
+		podEventNotificationChan: podEventNotificationChan,
+		log:                      log,
+	}
+	p.dataStore = cache.NewIndexer(cache.MetaNamespaceKeyFunc, NamespaceKeyIndexerFunc())
```
M00nF1sh, line 94: p.dataStore = cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{k8s.NamespaceIndexKey: k8s.NamespaceIndexFunc})

pkg/k8s/pod_wrapper.go
```diff
@@ -0,0 +1,94 @@
...
+	if opts.Namespace != "" {
+		items, err = k.podController.GetDataStore().ByIndex(NamespaceIndexKey, opts.Namespace)
+	} else {
+		items = k.podController.GetDataStore().List()
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	if err != nil {
```
M00nF1sh, line 65: two redundent check

pkg/k8s/custom_controller.go
```diff
@@ -0,0 +1,250 @@
...
+	c.podEventNotificationChan <- GenericEvent{
+		EventType: CREATE,
+		Meta:      meta,
+		Object:    obj.(runtime.Object),
+	}
+	return nil
+}
+
+// notifyChannelOnCreate notifies the add event on the appropriate channel
+func (c *PodController) notifyChannelOnUpdate(oldObj, newObj interface{}) error {
```
M00nF1sh, line 217: the error here seems not be used anywhere, at least we should have logs.

pkg/conversions/pod_conversion.go
cgchinmay: I feel internally we do need to store resource and resourceType. So Instead I made them private and changed them name so that only constructor initializes with predefined values

pkg/conversions/pod_conversion.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, fmt.Errorf("failed to convert object to pod")
+	}
+	return c.stripDownPod(pod), nil
+}
+
+// ConvertList converts the original pod list to stripped down list of pod objects
+func (c *podConverter) ConvertList(originalList interface{}) (convertedList interface{}, err error) {
+	podList, ok := originalList.(*v1.PodList)
+	if !ok {
+		return nil, fmt.Errorf("faield to convert object to pod list")
```
M00nF1sh, line 53: failed

pkg/conversions/pod_conversion_test.go
```diff
@@ -0,0 +1,102 @@
...
+			*pod1,
+			*pod2,
+		},
+	}
+
+	convertedList, err := podConverter.ConvertList(podList)
+	if err != nil {
+		t.Error(err)
+	}
+	assert.NotNil(t, convertedList, "Converted List cannot be Nil")
```
M00nF1sh, line 100: can we also verify the convertedList indeed stripped down

pkg/k8s/custom_controller.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	c.controller = cache.New(config)
+
+	// Run the controller
+	c.controller.Run(stopChanel)
+}
+
+// GetDataStore returns the data store when it has successfully synced with API Server
+func (c *CustomController) GetDataStore() cache.Indexer {
+	// Custom data store, it should not be accessed directly as the cache could be out of sync
```
M00nF1sh, line 157: nit, i'm ok with this implementation as it's already used by vpc resource controller. 
However, we should refactor this in the future, as this approach will make controllers to run without have pod synced. (it actually blocks when pod information is accessed).

pkg/k8s/custom_source.go
```diff
@@ -0,0 +1,174 @@
...
+
+// NotificationChannel monitors channels of type Create/Update/Delete
+type NotificationChannel struct {
+	// once ensures the event distribution goroutine will be performed only once
+	once sync.Once
+
+	// stop is to end ongoing goroutine, and close the Create channel
+	stop <-chan struct{}
+
+	Source <-chan GenericEvent
```
M00nF1sh, line 63: this should be `source`

pkg/k8s/custom_source.go
```diff
@@ -0,0 +1,174 @@
...
+
+	// dest is the destination channels of the Pod event handlers
+	dest []chan GenericEvent
+
+	// DestBufferSize is the specified buffer size of dest channels.
+	// Default to 1024 if not specified.
+	DestBufferSize int
```
M00nF1sh, line 70: `destBufferSize`

pkg/k8s/pod_wrapper.go
```diff
@@ -0,0 +1,91 @@
...
+// PodsRepository represents an interface with all the common operations on pod objects
+type PodsRepository interface {
+	GetPod(namespace string, name string) (*v1.Pod, error)
+	ListPodsWithMatchingLabels(opts client.ListOptions) (*v1.PodList, error)
+}
+
+// podsRepository is the wrapper object with the client
+type podsRepository struct {
+	customController *CustomController
+	cacheClient      client.Client
```
M00nF1sh, line 22: `cacheClient` is not used, let's remove it

pkg/k8s/custom_source.go
cgchinmay: It is used in cloudmap_controller to set the Source. It is an exported field

main.go
```diff
@@ -56,6 +62,8 @@ import (
 const (
 	flagHealthProbeBindAddr       = "health-probe-bind-addr"
 	defaultHealthProbeBindAddress = ":61779"
+	defaultAPIServerQPS           = 1e6
```
abhipth, line 65: Is there a reason why we are adding such high default QPS? I believe this could negatively impact other components including API Server if this controller is misbehaving for some reason.

main.go
```diff
@@ -123,6 +144,19 @@ func main() {
...
 		HealthProbeBindAddress: healthProbeBindAddress,
 	})
+
+	converter := conversions.NewPodConverter()
+
+	customController := k8s.NewCustomController(
+		clientSet,
+		listPageLimit,
+		metav1.NamespaceAll,
+		converter,
```
abhipth, line 154: nit: using conversions.NewPodConverter() directly would be cleaner I think. But confirm with Yang once.

pkg/conversions/pod_conversion_test.go
```diff
@@ -0,0 +1,102 @@
...
+		},
+		Spec: v1.PodSpec{
+			NodeName:   "TestNode",
+			Containers: containers,
+		},
+	}
+
+	convertedObj, err := podConverter.ConvertObject(oldPod)
+	if err != nil {
+		t.Error(err)
```
abhipth, line 50: Can you check error assert statements, they may be better for checking if there is no error.
https://godoc.org/github.com/stretchr/testify/assert

pkg/k8s/custom_controller.go
abhipth: +1 I will take an action item to see if we can refactor to sync the cache first and then start the controllers.

pkg/k8s/custom_source.go
```diff
@@ -0,0 +1,174 @@
...
+	for _, dst := range cs.dest {
+		// We cannot make it under goroutine here, or we'll meet the
+		// race condition of writing message to closed channels.
+		// To avoid blocking, the dest channels are expected to be of
+		// proper buffer size. If we still see it blocked, then
+		// the controller is thought to be in an abnormal state.
+		dst <- evt
+	}
+}
+
```
abhipth, line 163: Can we add description of each function in comments so that it's easy to understand what each function is doing.

main.go
cgchinmay: These values are from from the loadbalancer repo

M00nF1sh (approved):

M00nF1sh (approved):

Issue #425, opened 2021-01-21
Allow configuring tracing address as status.hostIP using the downward API


fawadkhaliq:
**Is your request related to a problem?**
Some of the tracing providers support running agents as daemonsets (e.g. DataDog). If the agent is running on node locally and bound to the node local IP address, the tracing agent IP address will be different for every Envoy. AppMesh controller takes a global endpoint that is supposed to be same for every Envoy (either a service endpoint or a localhost for sidecar tracing agents). This works well for agents running at cluster level or as sidecars. However, anything running as a daemonset will not be able to receive traces. 

**Describe the solution you'd like**
For tracing agents running as daemonsets, appmesh controller should provide a way to configure the host IP using fieldRef, which can be rendered using the [Downward API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#capabilities-of-the-downward-api). The downward API supports `status.hostIP`

**Describe alternatives you've considered**
Have the tracing agent support the k8s service topology to expose node local endpoints. This change is not on the controller side but on how the tracing agents are deployed so there's not much App Mesh controller can control

fawadkhaliq:
Fixed here: https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/426

PR #426, opened 2021-01-22
Configure tracing address as the downward API ref to status.hostIP

fawadkhaliq:
**Issue** #425 

**Description of changes**
Some of the tracing providers support running agents as daemonsets (e.g. DataDog). If the agent is running on node locally and bound to the node local IP address, the tracing agent IP address will be different for every Envoy. AppMesh controller takes a global endpoint that is supposed to be same for every Envoy (either a service endpoint or a localhost for sidecar tracing agents). This works well for agents running at cluster level or as sidecars. However, anything running as a daemonset will not be able to receive traces.

This change adds support to use the k8s downward API to reference the pod `status.hostIP`.  For daemonsets running on host network namespace, `status.hostIP` Envoy will be able to send traces to them. 

Also, removed the fragile approach of text/template rendering of the Envoy container spec string JSON we were using and moved into the sidecar builder using proper container type. 

**Usage/example**
e.g. Enable DataDog tracing and point Envoy sidecars to their node local DataDog daemonsets running on host network namespace. You will need to set the following flags and values at controller init:

```
--enable-datadog-tracing=true
--datadog-port=8126
--datadog-addres=ref:status.hostIP
```


pkg/inject/envoy.go
```diff
@@ -187,15 +86,7 @@ func (m *envoyMutator) mutate(pod *corev1.Pod) error {
...
-	envoySidecar, err := renderTemplate("envoy", envoyContainerTemplate, variables)
-	if err != nil {
-		return err
-	}
-	container := corev1.Container{}
-	err = json.Unmarshal([]byte(envoySidecar), &container)
-	if err != nil {
-		return err
-	}
+	container := buildEnvoySidecar(variables)
```
achevuru, line 89: Like the idea. However, if we want to move away from templates then I think we should change how we currently inject envoys in to VirtualGateways as well so that we have uniformity on how we inject envoys - https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/inject/virtualgateway_envoy.go#L14

achevuru:
Moving away from templates will run in to a conflict with the current mTLS merge PR. Might be better, we merge that first and add support for the new SDS_UDS_PATH env variable in the current flow as well. Should be a minor change on top of this diff.

pkg/inject/envoy.go
achevuru: Also, we had a  brief discussion around this in the past but might be worth adding datadog support for VirtualGateways as well now that we are making a change around this area. Right now we only have support for XrayTracing in VirtualGateways.

fawadkhaliq:
> Moving away from templates will run in to a conflict with the current mTLS merge PR. Might be better, we merge that first and add support for the new SDS_UDS_PATH env variable in the current flow as well. Should be a minor change on top of this diff.

Sounds good. Will rebase on top of mTLS PR

pkg/inject/envoy.go
fawadkhaliq: I'd break those into separate PRs. Virtual gateway is similar but not similar (we "optionally" Envoy and it's config). For the DataDog tracers in Virtual gateways, it should definitely be a separate PR as users have control over virtual gateway envoy spec until virtual nodes

pkg/inject/envoy.go
```diff
@@ -18,6 +18,8 @@ type EnvoyTemplateVariables struct {
 	MeshName                     string
 	VirtualNodeName              string
 	Preview                      string
+	EnableSDS                    bool
```
achevuru: Interesting, this is showing up as diff, when this change is already in master. 
https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/inject/envoy.go#L126

fawadkhaliq: Check the PR instead of the last commit - that will show the accurate diff to the main branch

achevuru (approved):

PR #445, opened 2021-02-25
Bump Envoy image version to v1.16.1.1-prod

flashyang:
*Issue #, if available:*
https://github.com/aws/aws-app-mesh-roadmap/issues/315

*Description of changes:*
Bump AppMesh Envoy image version from v1.16.1.0-prod to v1.16.1.1-prod

The new Envoy image contains fix for the following issues and bugs:
1. Envoy stay in a stale connection and never reconnect to ADS server after ADS server hard reboot: https://github.com/envoyproxy/envoy/issues/5173 
2. Missing URL information when inspecting traces in the Datadog UI: https://github.com/aws/aws-app-mesh-roadmap/issues/309
3. Envoy Mangles RequestId when Tracing is Enabled: https://github.com/envoyproxy/envoy/issues/11532

### DIFF
--- a/config/helm/appmesh-controller/test.yaml
+++ b/config/helm/appmesh-controller/test.yaml
@@ -15,7 +15,7 @@ image:
 sidecar:
   image:
     repository: 840364872350.dkr.ecr.us-west-2.amazonaws.com/aws-appmesh-envoy
-    tag: v1.16.1.0-prod
+    tag: v1.16.1.1-prod
     # sidecar.logLevel: Envoy log level can be info, warn, error or debug
   logLevel: info
   envoyAdminAccessPort: 9901
--- a/config/helm/appmesh-controller/values.yaml
+++ b/config/helm/appmesh-controller/values.yaml
@@ -15,7 +15,7 @@ image:
 sidecar:
   image:
     repository: 840364872350.dkr.ecr.us-west-2.amazonaws.com/aws-appmesh-envoy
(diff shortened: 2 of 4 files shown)


fawadkhaliq:
Although Helm chart test file [1] is only used for linting and chart gen, it's good to update the version there too

[1] https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/config/helm/appmesh-controller/test.yaml#L18

fawadkhaliq:
> Envoy Mangles RequestId when Tracing is Enabled: envoyproxy/envoy#11532

It seems the the envoy issue https://github.com/envoyproxy/envoy/issues/11532 is still open and the corresponding PR https://github.com/envoyproxy/envoy/pull/14752 was closed (to be revisited). Wondering what the fix includes in this patch release?

fawadkhaliq (approved):

lavignes:
@fawadkhaliq Actually I think that was a bit premature! We had added some code to mitigate that issue in this release (so it got included in the list for sake of completion) but we have not enabled it. Right now there is no way to turn this mitigation on so I think we can safely say that this does not in fact fix that issue yet. :)

We're still investigating the best way to control this behavior and will create a new issue on our roadmap to track it specifically from our end. We're be a bit more diligent on these change lists in the future. I would prefer we link them back to App Mesh roadmap issues to better contextualize the problem and document the workarounds/fixes.

PR #458, opened 2021-03-24
Updated Readme for appmesh-controller Installation via helm

cgchinmay:
https://github.com/aws/eks-charts/issues/477

*Description of changes:*
Updated IRSA instructions in appmesh-controller readme via helm 

### DIFF
--- a/config/helm/appmesh-controller/README.md
+++ b/config/helm/appmesh-controller/README.md
@@ -31,8 +31,7 @@ Create namespace
 kubectl create ns appmesh-system
 ```
 
-The controller runs on the worker nodes, so it needs access to the AWS App Mesh / Cloud Map resources via IAM permissions. The
-IAM permissions can either be setup via IAM roles for service account or can be attached directly to the worker node IAM roles.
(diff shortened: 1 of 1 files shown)


config/helm/appmesh-controller/README.md
```diff
@@ -89,18 +89,49 @@ helm upgrade -i appmesh-controller eks/appmesh-controller \
...
+Create an IAM policy called AWSAppMeshEnvoyIAMPolicy
+```
+aws iam create-policy \
+    --policy-name AWSAppMeshEnvoyIAMPolicy \
+    --policy-document file://envoy-iam-policy.json
+```
+
+Take note of the policy ARN that is returned
+
+Create an IAM role for service account for the App Mesh Envoy, use the ARN from the step above. Ensure that Application Namespace already exists
```
achevuru, line 112: We can also add that if the customer's Mesh enabled applications are already using IRSA then we can add the above envoy specific policies to the same policy.

cgchinmay: Will add a note specifying that Trust Relationship should be edited so that IAM role for the IRSA can be shared across namespaces

achevuru (approved):

Issue #460, opened 2021-03-26
VirtualGateway does not have tracing enabled when using jaeger but does with aws xray


caleygoff-invitae:
**Describe the bug**
We are exploring tracing signals through the mesh and are comparing Jaeger with XRay. I discovered that the VirtualGateway doesn't have `tracing` enabled when using Jaeger on the appmesh-controller and does when using AWS XRay.

**Steps to reproduce**
1) Install Appmesh controller
2) Enable `xray` on the appmesh controller
```
        - --enable-xray-tracing=true
        - --xray-image=amazon/aws-xray-daemon:latest
        - --xray-daemon-port=2000
```
3) Deploy a `VirtualGateway`
4) Exec into the VirtualGateways Envoy container and look at the envoy config file at `/tmp`. Observe the following:
```
staticResources:
  clusters:
  - connectTimeout: 1s
    http2ProtocolOptions: {}
    loadAssignment:
      clusterName: static_cluster_sds_unix_socket
      endpoints:
      - lbEndpoints:
        - endpoint:
            address:
              pipe:
                path: /run/spire/sockets/agent.sock
    name: static_cluster_sds_unix_socket
    type: STATIC
tracing:
  http:
    config:
      daemon_endpoint:
        address: 127.0.0.1
        port_value: 2000
        protocol: UDP
      segment_fields:
        aws:
          app_mesh:
            mesh_name: dev
            virtual_node_name: mesh-ingress-gateway-single-cluster-test_single-cluster-test
        origin: AWS::AppMesh::Proxy
      segment_name: dev/mesh-ingress-gateway-single-cluster-test_single-cluster-test
    name: envoy.tracers.xray
```
5) update  the appmesh controller to use the `jaeger` tracing backend 
```
        - --enable-jaeger-tracing=true
        - --jaeger-address=jaeger-collector.system
        - --jaeger-port=80
```
6) Deploy a `VirtualGateway`
7) Exec into the VirtualGateways `Envoy` container and look at the envoy config file at `/tmp`. Observe the following:
```
staticResources:
  clusters:
  - connectTimeout: 1s
    http2ProtocolOptions: {}
    loadAssignment:
      clusterName: static_cluster_sds_unix_socket
      endpoints:
      - lbEndpoints:
        - endpoint:
            address:
              pipe:
                path: /run/spire/sockets/agent.sock
    name: static_cluster_sds_unix_socket
    type: STATIC
```
**Expected outcome**
[...]

PR #462, opened 2021-04-07
Enable Jaeger tracing for VirtualGateway Envoy

cgchinmay:
*Issue #460 *

*Description of changes:*
Enabled jaeger tracing for Virtual Gateway Envoy

Testing Result: Inspected Envoy config in Virtual Gateway Envoy container  as /tmp/envoy/envoyconf.yaml
```
tracing:
 http:
  name: envoy.tracers.zipkin
  typed_config:
   "@type": type.googleapis.com/envoy.config.trace.v2.ZipkinConfig
   collector_cluster: jaeger
   collector_endpoint: "/api/v2/spans"
   collector_endpoint_version: HTTP_JSON
   shared_span_context: false
static_resources:
  clusters:
  - name: jaeger
    connect_timeout: 1s
    type: STRICT_DNS
    lb_policy: ROUND_ROBIN
    load_assignment:
      cluster_name: jaeger
      endpoints:
      - lb_endpoints:
        - endpoint:
           address:
            socket_address:
             address: jaeger-collector.system
             port_value: 80
```

### DIFF
--- a/pkg/inject/virtualgateway_envoy.go
+++ b/pkg/inject/virtualgateway_envoy.go
@@ -45,8 +45,8 @@ type virtualGatwayEnvoyConfig struct {
 	enableXrayTracing          bool
 	xrayDaemonPort             int32
 	enableJaegerTracing        bool
-	jaegerPort                 string
 	jaegerAddress              string
+	jaegerPort                 int32
 }
 
 // newVirtualGatewayEnvoyConfig constructs new newVirtualGatewayEnvoyConfig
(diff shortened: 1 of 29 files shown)


pkg/inject/virtualgateway_envoy.go
```diff
@@ -20,22 +21,26 @@ const envoyVirtualGatewayEnvMap = `
   "ENVOY_ADMIN_ACCESS_LOG_FILE": "{{ .AdminAccessLogFile }}",
   "AWS_REGION": "{{ .AWSRegion }}"{{ if .EnableSDS }},
   "APPMESH_SDS_SOCKET_PATH": "{{ .SdsUdsPath }}"{{ end }}{{ if .EnableXrayTracing }},
-  "ENABLE_ENVOY_XRAY_TRACING": "1","XRAY_DAEMON_PORT": "{{ .XrayDaemonPort }}"{{ end }}
+  "ENABLE_ENVOY_XRAY_TRACING": "1","XRAY_DAEMON_PORT": "{{ .XrayDaemonPort }}"{{ end }}{{ if .EnableJaegerTracing }},
+  "ENABLE_ENVOY_JAEGER_TRACING": "1","JAEGER_TRACER_ADDRESS": "{{ .JaegerAddress }}","JAEGER_TRACER_PORT": "{{ .JaegerPort }}"{{ end }}
```
lavignes, line 25: Just driving by this PR to say thanks for doing this!

I have a comment regarding this implementation vs the way we do it for virtual nodes. It seems like for virtual nodes we are generating a JSON config file with raw Envoy config for the tracer while for virtual gateways we simply pass the environment variables to the App Mesh envoy image.

Would it be very difficult to include making it consistent in this PR? Ideally we should just pass the `ENVOY_JAEGAR*` environment variables for both types of resources.

Coincidentally I was planning on making a PR to do this soon, hence why I'm asking now.

cgchinmay: Thanks for the suggestion. I noticed the existing implementation for x-ray, so followed the same pattern to keep my change minimum. I will take a look at the virtual node approach and accordingly update this PR.

lavignes: I was actually advocating to _just use the environment variables_ in the Envoy instead of mounting the envoyconf.yaml. I was planning to send a PR to remove this config mounting since the Envoy API can break between versions but the image can always generate the correct config for its given Envoy version.

cgchinmay: @lavignes Thanks for clarifying. In that case I will keep this change limited to virtual gateway envoy itself, will pass the environment variables instead of mounting the config file. 

I will have another PR soon to remove the mounting logic from virtual node as that would require testing as well.
Only for testing purpose and scope, i think it's better to have them separate.

lavignes: Ack. Sounds good to me

Issue #450, opened 2021-03-03
Unable to reference Virtual Gateway in GatewayRoute definition


vodwood:
**Describe the bug**
If I am not mistaken, there is no way currently to assign a GatewayRoute to a specific VirtualGateway via CRD, while it is possible via App Mesh User Interface.

**Steps to reproduce**
1. Create namespace and mesh
2. In this namespace, create two virtual gateways with two correlating standalone envoy deployments
3. Create two gateway routes for separate applications and apply
4. Receive: "Error from server (found multiple matching virtualGateways for namespace: namespace, expecting 1 but found N"*
* Proceeds to list virtual gateways in all namespaces, even though query was kubectl -n namespace get virtualgateways

**Expected outcome**
As via UI, there needs to be a method of mapping gateway routes to specific gateways via CRD, because there might be more than one gateway in a namespace

**Environment**

* App Mesh controller version: appmesh-controller:v1.3.0
* Envoy version: aws-appmesh-envoy:v1.16.1.1-prod 
* Are you using any integrations? X-ray, Jaeger etc. If so versions? aws-xray-daemon:latest
* Kubernetes version 1.18
* Using EKS (yes/no), if so version? yes, eks.3

**Additional Context**:
This is a fairly standard use case - to allow integration testing of every application, I'm exposing every application on the mesh via a dedicated gateway, since host-based routing is not available for ingress yet.

Currently creating Service, Deployment of envoy and VirtualGateway via CRD and adding route manually via UI and it works perfectly - each gateway routes to each application that needs to be exposed outside.

cgchinmay:
When creating Virtual Gateway and GatewayRoutes via CRD, this is the logic we have -> The virtual gateway selects namespaces using the namespaceSelector, here you can select multiple namespaces. But from namespace point of view, it is expected to have only 1 virtual gateway and all gateway routes defined in that namespace get attached to that 1 virtual gateway. In CRD's we don't add any virtual gateway reference while creating gateway route as this association is done at the namespace level. So in your case you have 2 virtual gateways pointing to one namespace and thats why the error. We would update our docs to clarify this. Here's a sample walkthrough to see Virtual Gateway Usage: 
`
https://github.com/aws/aws-app-mesh-examples/tree/main/walkthroughs/howto-k8s-ingress-gateway
`

vodwood:
[...]

PR #464, opened 2021-04-09
Added a new GatewayRoute selector optional field to VirtualGateway

cgchinmay:
Updated membership_designator logic for GatewayRoute to VirtualGateway

*Issue #450 

*Description of changes:*
Before Change:
VirtualGateway would select GatewayRoutes based on the namespaceSelector

After Change:
In addition to namespace selector, now VirtualGateway can specify gatewayRouteSelector to further select a subset of GatewayRoutes. 

Why is this Change Needed ?
Now customers can have VirtualGateway A which would select subset of GatewayRoutes in namespace X and have VirtualGateway B which would select other subset of GatewayRoutes in the same namespace X based on the gatewayRouteSelector labels. 

Sample Spec
```
kind: VirtualGateway
metadata:
  name: ingress-gw
  namespace: ${APP_NAMESPACE}
spec:
  namespaceSelector:
    matchLabels:
      gateway: ingress-gw
  gatewayRouteSelector:
    matchLabels:
      gateway: ingress-gw
  podSelector:
    matchLabels:
      app: ingress-gw
  listeners:
    - portMapping:
        port: 8088
        protocol: http
---
kind: GatewayRoute
metadata:
  name: gateway-route-headers
  namespace: ${APP_NAMESPACE}
spec:
  httpRoute:
    match:
      prefix: "/headers"
    action:
      target:
        virtualService:
          virtualServiceRef:
            name: color-headers
---
kind: GatewayRoute
metadata:
  name: gateway-route-paths
  namespace: ${APP_NAMESPACE}
  labels:
    gateway: ingress-gw
spec:
  httpRoute:
    match:
      prefix: "/paths"
    action:
      target:
        virtualService:
          virtualServiceRef:
            name: color-paths
----
```
For above Spec, the VirtualGateway will match with only 1 GatewayRoute: gateway-route-paths as it has the matching gatewayRouteSelector. 

For GatewayRoute: gateway-route-headers, there won't be any matching VirtualGateways, so the user will see following error
```
failed to find matching virtualGateway for gatewayRoute: gateway-route-headers, expecting 1 but found 0)
```

### DIFF
--- a/apis/appmesh/v1beta2/virtualgateway_types.go
+++ b/apis/appmesh/v1beta2/virtualgateway_types.go
@@ -322,6 +322,10 @@ type VirtualGatewaySpec struct {
 	// 	if absent, it selects no pod.
 	// +optional
 	PodSelector *metav1.LabelSelector `json:"podSelector,omitempty"`
+	// GatewayRouteSelector selects GatewayRoutes using labels to designate GatewayRout membership.
+	// This field follows standard label selector semantics; if present but empty, it selects all in that namespace.
+	// +optional
+	GatewayRouteSelector *metav1.LabelSelector `json:"gatewayrouteSelector,omitempty"`
 	// The listener that the virtual gateway is expected to receive inbound traffic from
 	// +kubebuilder:validation:MinItems=0
 	// +kubebuilder:validation:MaxItems=1
--- a/apis/appmesh/v1beta2/virtualgateway_types.go
+++ b/apis/appmesh/v1beta2/virtualgateway_types.go
@@ -322,7 +322,7 @@ type VirtualGatewaySpec struct {
 	// 	if absent, it selects no pod.
 	// +optional
 	PodSelector *metav1.LabelSelector `json:"podSelector,omitempty"`
-	// GatewayRouteSelector selects GatewayRoutes using labels to designate GatewayRout membership.
+	// GatewayRouteSelector selects GatewayRoutes using labels to designate GatewayRoute membership.
 	// If not specified it selects all GatewayRoutes in that namespace.
 	// +optional
 	GatewayRouteSelector *metav1.LabelSelector `json:"gatewayRouteSelector,omitempty"`
--- a/apis/appmesh/v1beta2/virtualgateway_types.go
+++ b/apis/appmesh/v1beta2/virtualgateway_types.go
@@ -323,9 +323,9 @@ type VirtualGatewaySpec struct {
 	// +optional
 	PodSelector *metav1.LabelSelector `json:"podSelector,omitempty"`
 	// GatewayRouteSelector selects GatewayRoutes using labels to designate GatewayRout membership.
-	// This field follows standard label selector semantics; if present but empty, it selects all in that namespace.
+	// If not specified it selects all GatewayRoutes in that namespace.
 	// +optional
-	GatewayRouteSelector *metav1.LabelSelector `json:"gatewayrouteSelector,omitempty"`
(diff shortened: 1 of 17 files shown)


apis/appmesh/v1beta2/virtualgateway_types.go
```diff
@@ -322,6 +322,10 @@ type VirtualGatewaySpec struct {
 	// 	if absent, it selects no pod.
 	// +optional
 	PodSelector *metav1.LabelSelector `json:"podSelector,omitempty"`
+	// GatewayRouteSelector selects GatewayRoutes using labels to designate GatewayRout membership.
+	// This field follows standard label selector semantics; if present but empty, it selects all in that namespace.
+	// +optional
+	GatewayRouteSelector *metav1.LabelSelector `json:"gatewayrouteSelector,omitempty"`
 	// The listener that the virtual gateway is expected to receive inbound traffic from
```
M00nF1sh, line 329: the json field should be `gatewayRouteSelector`

apis/appmesh/v1beta2/virtualgateway_types.go
```diff
@@ -322,6 +322,10 @@ type VirtualGatewaySpec struct {
 	// 	if absent, it selects no pod.
 	// +optional
 	PodSelector *metav1.LabelSelector `json:"podSelector,omitempty"`
+	// GatewayRouteSelector selects GatewayRoutes using labels to designate GatewayRout membership.
+	// This field follows standard label selector semantics; if present but empty, it selects all in that namespace.
```
M00nF1sh, line 326: we need to mention that `if it's unspecified or present but empty, it selects all gatewayRoute in that namespace`

apis/appmesh/v1beta2/virtualgateway_types.go
```diff
@@ -322,6 +322,10 @@ type VirtualGatewaySpec struct {
 	// 	if absent, it selects no pod.
 	// +optional
 	PodSelector *metav1.LabelSelector `json:"podSelector,omitempty"`
+	// GatewayRouteSelector selects GatewayRoutes using labels to designate GatewayRout membership.
```
M00nF1sh, line 325: s/GatewayRout/GatewayRoute

M00nF1sh (approved):
/lgtm

Also, it's better to include more details in the description. e.g. our internal design docs(exclude any internal links) to give more public visibility.

M00nF1sh (approved):
approved

M00nF1sh:
One reminder: don't update the crd in eks-charts yet. let's first update the instructions in AWS Docs and https://github.com/aws/eks-charts/tree/master/stable/appmesh-controller
so instead of 
```kubectl apply -k "github.com/aws/eks-charts/stable/appmesh-controller//crds?ref=master"```
use a ref to a released version/tag

File: pkg/virtualgateway/membership_designator_test.go [hunk 4]
```diff
@@ -472,6 +491,39 @@ func Test_virtualGatewayMembershipDesignator_DesignateForGatewayRoute(t *testing
 			want:    nil,
 			wantErr: errors.New("failed to find matching virtualGateway for namespace: awesome-ns, expecting 1 but found 0"),
 		},
+		{
+			name: "virtualgateway with valid gatewayroute selector",
```
achevuru, line 671: How about test cases for multiple and no match cases? Do we have them already?

achevuru: I see them in the new diff..

PR #468, opened 2021-04-19
Update jaegar tracing Envoy config to v3 API

lavignes:
*Description of changes:*

The xDS v2 API is no longer supported in Envoy 1.18 and has been deprecated for some time. App Mesh will be defaulting Envoy to the xDS v3 API in our 1.17 image so I'd like to update the version for the Jaegar tracer config to v3 at least so we may support Envoy 1.18.

The ideal solution is to use the `ENABLE_ENVOY_JAEGER_TRACING` and [related environment variables](https://docs.aws.amazon.com/app-mesh/latest/userguide/envoy-config.html) and let the Envoy image generate the config like is being done for Virtual Gateways in #462.

But I want to make sure we put this minimal change in place to ensure we stay compatible for the time being.

### DIFF
--- a/pkg/inject/jaeger.go
+++ b/pkg/inject/jaeger.go
@@ -10,7 +10,7 @@ tracing:
  http:
   name: envoy.tracers.zipkin
   typed_config:
-   "@type": type.googleapis.com/envoy.config.trace.v2.ZipkinConfig
+   "@type": type.googleapis.com/envoy.config.trace.v3.ZipkinConfig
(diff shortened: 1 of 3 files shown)


fawadkhaliq (approved):

fawadkhaliq:
>The ideal solution is to use the ENABLE_ENVOY_JAEGER_TRACING

++, we should move to ENABLE_ENVOY_JAEGER_TRACING instead

cc: @cgchinmay

PR #479, opened 2021-06-10
add responseType to configure DNS service discovery type

fawadkhaliq:
**Issue #, if available:**
None

**Description of changes**
Added the support to configure strict and logical DNS service discovery types in the Envoy sidecars. 

When using **strict DNS** service discovery, Envoy will continuously and asynchronously resolve the specified DNS targets. Each returned IP address in the DNS result will be considered an explicit host in the upstream cluster. This means that if the query returns three IP addresses, Envoy will assume the cluster has three hosts, and all three should be load balanced to. If a host is removed from the result Envoy assumes it no longer exists and will drain traffic from any existing connection pools. **Logical DNS** uses a similar asynchronous resolution mechanism to strict DNS. However, instead of strictly taking the results of the DNS query and assuming that they comprise the entire upstream cluster, a logical DNS cluster only uses the first IP address returned when a new connection needs to be initiated.

This PR exposed the strict vs logical DNS using a `responseType` field in the virtual nodes. `responseType` is an optional field that supports two values: `ENDPOINTS` and `LOADBALANCER`. `ENDPOINTS` maps to strict DNS mode in Envoy and `LOADBALANCER` maps to logical DNS mode in Envoy. If `responseType` is not set, the default value is set to `LOADBALANCER` (i.e. logical DNS, the existing behavior) 

```
kind: VirtualNode
metadata:
  name: vn1
  namespace: ns1
spec:
  podSelector:
    matchLabels:
      app: app1
      version: v1
  listeners:
    - portMapping:
        port: 8080
        protocol: http
  serviceDiscovery:
    dns:
      hostname: svc1.ns1.svc.cluster.local
      responseType: ENDPOINTS
```
When using ENDPOINTS(strict DNS)  `responseType` in Kubernetes, you will need to expose the Kubernetes services in a headless manner. Here’s a sample headless Service (note ClusterIP: None)

```
apiVersion: v1
kind: Service
metadata:
  name: svc1
  namespace: ns1
spec:
  clusterIP: None
  ports:
    - name: http
      port: 8080
  selector:
    app: app1
  type: ClusterIP
```


apis/appmesh/v1beta2/virtualnode_types.go
```diff
@@ -331,6 +331,9 @@ type AWSCloudMapServiceDiscovery struct {
 type DNSServiceDiscovery struct {
 	// Specifies the DNS service discovery hostname for the virtual node.
 	Hostname string `json:"hostname"`
+	// Choose between ENDPOINTS (strict DNS) and LOADBALANCER (logical DNS) mode in Envoy sidecar
+	// +optional
+	ResponseType *string `json:"responseType,omitempty"`
```
cgchinmay, line 336: Can we add this validation here ? 
// +kubebuilder:validation:Enum=ENDPOINTS;LOADBALANCER;

cgchinmay (approved):
just 1 comment, rest looks good

apis/appmesh/v1beta2/virtualnode_types.go
fawadkhaliq: ++ added!

achevuru (approved):

PR #487, opened 2021-06-25
Remove field length constraints from query parameter exact match

cgchinmay:
Updated Chart minor version

*Description of changes:*
Remove field length constraint for query parameter exact match 
reference: https://docs.aws.amazon.com/app-mesh/latest/APIReference/API_QueryParameterMatch.html


M00nF1sh (commented):
why is it only removed in types annotation? how about the crd yamls?

M00nF1sh (approved):
/lgtm

Issue #357, opened 2020-10-07
Allow service instance port to be specified in Cloud Map service discovery


rifqialabqary:
I'm trying to expose EKS services via Amazon API Gateway (HTTP API) through private integration using AWS Cloud Map service discovery.

If I understand correctly, through the App Mesh Controller, only the pods IP addresses will be registered.
However, API Gateway needs both IP addresses and ports in order to work when using AWS Cloud Map private integration as stated [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-private.html#http-api-develop-integrations-private-Cloud-Map).

Using the following, I am able to register both the IP addresses and ports in AWS Cloud Map. And I am able to connect to the EKS pods through API Gateway.

```
serviceDiscovery:
  awsCloudMap:
    namespaceName: ${CLOUDMAP_NAMESPACE}
    serviceName: ${CLOUDMAP_SERVICE}
    attributes:
    - key: "AWS_INSTANCE_PORT"
      value: "${APP_PORT}"
```

However, although Custom Resources in Kubernetes are created, resources in App Mesh are not being created and logs in the controller shows the following:

```
{
  "level": "error",
  "ts": 1601945565.5024025,
  "logger": "controller-runtime.controller",
  "msg": "Reconciler error",
  "controller": "virtualnode",
  "request": "eems-coe/eems-auth-service",
  "error": "BadRequestException: AWS CloudMap Service Discovery Attribute Key AWS_INSTANCE_PORT can not use the reserved AWS_ prefix.",
[...]

PR #490, opened 2021-07-16
Add service instance port to Cloud Map Service Discovery

Shreya027:
Issue #357

- Added instance port to Cloud Map Custom Attributes
- Standardized const variable naming across initialization comments

### DIFF
--- a/pkg/cloudmap/instances_reconciler.go
+++ b/pkg/cloudmap/instances_reconciler.go
@@ -14,11 +14,15 @@ import (
 )
 
 const (
-	// attrAWSInstanceIPV4 is a special attribute expected by CloudMap.
+	// AttrAWSInstanceIPV4 is a special attribute expected by CloudMap.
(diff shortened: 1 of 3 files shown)


Shreya027:
Committed code with IT & UT. Added data type conversion for port number.

pkg/cloudmap/instances_reconciler.go
```diff
@@ -151,6 +155,9 @@ func (r *defaultInstancesReconciler) buildInstanceAttributes(ms *appmesh.Mesh, v
 	}
 	podsNodeName := pod.Spec.NodeName
 	attr[AttrAWSInstanceIPV4] = pod.Status.PodIP
+
+	//In future, in case of multiple port mappings, the first mapping will be selected.
```
achevuru, line 160: Might be better to reword the comment as - `VirtualNode currently supports only one listener but even if support for multiple listeners in introduced, we will always derive the port value from the first listener config`.

Shreya027: Implemented re-wording changes for comment.

achevuru (approved):

Issue #491, opened 2021-08-02
Upgrade CRDs to apiextensions.k8s.io/v1


allamand:
**Is your feature request related to a problem?**

When deploying the CRDs using this command (from Installing the chart):

```
kubectl apply -k "github.com/aws/eks-charts/stable/appmesh-controller//crds?ref=master"
```
I get the following warning:

```
W0719 13:06:33.130081   56372 warnings.go:70] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
```

**Describe the solution you'd like**
Upgrade the apiextensions CRD to the v1

cgchinmay:
@Shreya027  We do have a PR for this : https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/492

PR #492, opened 2021-08-02
bump to apiextensions v1

allamand:
Signed-off-by: Sebastien Allamand <sallaman@amazon.com>

*Issue #, if available:* #491

*Description of changes:*

Upgrade from `apiextensions.k8s.io/v1beta1` to `apiextensions.k8s.io/v1`

I wonder however how to upgrade helm chart for using the new CRD and to which version ? appreciate some help here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ IMAGE ?= $(REPO):$(VERSION)
 PREVIEW=false
 
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true,crdVersions=v1beta1"
+CRD_OPTIONS ?= "crd:trivialVersions=true,crdVersions=v1"
 
 # app mesh aws-sdk-go override in case we need to build against a custom version
 APPMESH_SDK_OVERRIDE ?= "n"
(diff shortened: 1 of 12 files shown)


cgchinmay:
Hi @allamand 
We noticed that this is a breaking change as v1 doesn't preserve unknown fields in the CRD. So we need to update CRD's to have preserve unknown field annotation.

cgchinmay (changes requested):
Need to resolve merge conflicts

Shreya027:
Hi, I think this PR also needs to take into consideration preservation of unknown fields that get dropped in CRD version v1. 

If you try adding preserveUnknownFields=true in the CRD_OPTIONS as shown below:

CRD_OPTIONS?="crd:trivialVersions=true,crdVersions=v1beta1,preserveUnknownFields=true" and apply the changes, you get the following error in the terminal:

                __you may only set PreserveUnknownFields to true with v1beta1 CRDs
                Error: not all generators ran successfully__
                
I figured out one option can be using` // +kubebuilder:pruning:PreserveUnknownFields` on top of type structs to preserve fields.

cgchinmay:
Closing this PR as we are tracking it here: https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/562

PR #498, opened 2021-08-13
Updated project dependencies

cgchinmay:
*Description of changes:*
Updated controller-runtime to v0.9.6
Updated helm to v3.6.1
Updated Add and Remove Finalizer logic, thanks to @kiranmeduri 
Refactored code to make it compatible with updated dependencies
Removed FieldMatchingFlags usage for Object Conversion methods
Specify leaderelectionLock to ConfigMapsResourceLock as the updated controller runtime doesn't set it by default and we get error in the controller logs if not set. 
Updated golang to v1.16 for github actions 

Testing:
Validated UTs ,e2e along with mTLS and Integration tests


cgchinmay:
Need to update go version for the build to succeed in github actions workflow

Dockerfile
```diff
@@ -20,6 +20,7 @@ RUN GIT_VERSION=$(git describe --tags --dirty --always) && \
 FROM amazonlinux:2
```
kiranmeduri, line 20: Are there any plans to move to https://gallery.ecr.aws/amazonlinux/amazonlinux ?

pkg/k8s/finalizers.go
```diff
@@ -38,34 +34,46 @@ type defaultFinalizerManager struct {
...
 
-func (m *defaultFinalizerManager) AddFinalizers(ctx context.Context, obj APIObject, finalizers ...string) error {
-	oldObj := obj.DeepCopyObject()
-	needsUpdate := false
-	for _, finalizer := range finalizers {
-		if !HasFinalizer(obj, finalizer) {
-			controllerutil.AddFinalizer(obj, finalizer)
-			needsUpdate = true
+func (m *defaultFinalizerManager) AddFinalizers(ctx context.Context, obj client.Object, finalizers ...string) error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
kiranmeduri, line 38: is there a unit-test to verify this new retry logic? I am assuming this is pretty standard snippet across other controller implementation. It may help to add a note on why retry is needed here. Thanks

kiranmeduri (commented):
I did cursory review, nothing jumps out other than some clarifying comments. Will leave it to @M00nF1sh to approve as he might have done similar work in alb controller.

pkg/k8s/finalizers.go
cgchinmay: We do have finalizers_test. Will check if we need to update it to verify the retry logic

Dockerfile
```diff
@@ -20,6 +20,7 @@ RUN GIT_VERSION=$(git describe --tags --dirty --always) && \
...
 RUN yum update -y && \
     yum install -y ca-certificates && \
+    yum update curl && \
```
M00nF1sh, line 23: why is curl needed?

main.go
```diff
@@ -255,13 +262,12 @@ func main() {
...
 	// Only start the controller when the leader election is won
-	mgr.Add(manager.RunnableFunc(func(stop <-chan struct{}) error {
+	mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
 		setupLog.Info("starting custom controller")
 
 		// Start the custom controller
-		customController.StartController(stop)
+		customController.StartController(ctx.Done())
 		// If the manager is stopped, signal the controller to stop as well.
-		<-stop
```
M00nF1sh, line 264: `<-stop` this should be kept here. otherwise, it will log "stopping the controller" immediately

pkg/conversions/pod_conversion.go
```diff
@@ -33,7 +32,7 @@ type podConverter struct {
 func NewPodConverter() *podConverter {
 	return &podConverter{
 		podResource:     "pods",
-		podResourceType: &corev1.Pod{},
+		podResourceType: &v1.Pod{},
```
M00nF1sh, line 35: can we keep it as corev1 given we also have metav1

Dockerfile
cgchinmay: It is used by amazonlinux:2 and needs an update to fix some of the security warnings

M00nF1sh (approved):
/lgtm

PR #502, opened 2021-08-23
Use root image of minimal EKD-D to read service account token

cgchinmay:
*Issue #, if available:*
Controller is unable to read service account token and fetch account id creds as the base image did not have root access

*Description of changes:*
Change the EKS-D base to root 


M00nF1sh (approved):
/lgtm

PR #510, opened 2021-09-08
Add nil check for cloudmap instanceport attribute

cgchinmay:
*Issue #, if available:*
Integration test for virtualnode with cloudmap were failing as value of nstancePort was nil 

*Description of changes:*
Added nil check to hand such case

### DIFF
--- a/test/integration/virtualnode/virtualnode_test.go
+++ b/test/integration/virtualnode/virtualnode_test.go
@@ -3,11 +3,12 @@ package virtualnode_test
 import (
 	"context"
 	"fmt"
+	"sync"
+	"time"
+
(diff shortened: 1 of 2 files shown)


test/integration/virtualnode/virtualnode_test.go
```diff
@@ -160,7 +161,7 @@ var _ = Describe("VirtualNode", func() {
 
 		})
 
-		It("should create a virtual node with CloudMap ServiceDiscovery enabled", func() {
+		FIt("should create a virtual node with CloudMap ServiceDiscovery enabled", func() {
```
achevuru, line 164: You might've to revert this change back..

achevuru (approved):

PR #511, opened 2021-09-08
revert the manual workflow change until we have it working

cgchinmay:
*Description of changes:*
Manual workflow status of Integration test is not getting reported even when we successfully trigger and complete it, so reverting this change until we have the working fix


cgchinmay:
not needed so closing it

PR #520, opened 2021-09-23
Use xray daemon from ecr public repo instead of dockerhub

suniltheta:
**Description of changes:**

Changed controller helm configuration to pull X-Ray daemon image from ECR public repo instead of Dockerhub.

Benefits:
1. Reduced deployment size
amazon/aws-xray-daemon latest image on Dockerhub is 62MB [link](https://hub.docker.com/r/amazon/aws-xray-daemon/tags?page=1&ordering=last_updated). But xray/aws-xray-daemon on ECR public repo which is built from scratch is just 3.51MB [link](https://gallery.ecr.aws/xray/aws-xray-daemon).
1. Won't get throttled for pulling images from Dockerhub

No functional change here.

&nbsp;

### DIFF
--- a/config/helm/appmesh-controller/values.yaml
+++ b/config/helm/appmesh-controller/values.yaml
@@ -43,7 +43,7 @@ init:
 
 xray:
   image:
-    repository: amazon/aws-xray-daemon
+    repository: public.ecr.aws/xray/aws-xray-daemon
     tag: latest
 
 nameOverride: ""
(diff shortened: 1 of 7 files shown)


config/helm/appmesh-controller/values.yaml
```diff
@@ -43,7 +43,7 @@ init:
 
 xray:
   image:
-    repository: amazon/aws-xray-daemon
+    repository: public.ecr.aws/xray/aws-xray-daemon
     tag: latest
```
cgchinmay, line 47: Can we change latest to the tag value instead?

cgchinmay (changes requested):
Can we change the latest to specific tag ?

config/helm/appmesh-controller/values.yaml
suniltheta: The latest docker image tag is still available on ECR public repo of aws-xray-daemon. Everytime they upload a new version they would update their latest I believe. So it is good to still keep this default tag to `latest` still?

cgchinmay: The reason of being specific is that, if a someone runs into an issue, we can quickly narrow it down.

suniltheta: As aws-xray-daemon is been stable for all these time, I think we can still keep the latest tag.

cgchinmay (approved):

PR #522, opened 2021-09-29
revert support for samplingRate

suniltheta:
**Issue**  
AppMesh roadmap https://github.com/aws/aws-app-mesh-roadmap/issues/371

**Description of changes:**  
Revert the changes to introduce the support for setting `XRAY_SAMPLING_RATE` via helm config samplingRate.

Previous change was done on PR https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/512 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #512, opened 2021-09-08
Introduce config variables samplingRate & statsSocketPath

suniltheta:
From v1.19.1.0-prod AppMesh Envoy image [release](https://github.com/aws/aws-app-mesh-roadmap/issues/369) we support two new env variables on Envoy proxy container.   
1. XRAY_SAMPLING_RATE
2. STATSD_SOCKET_PATH https://github.com/aws/aws-app-mesh-roadmap/issues/342

See docs: https://docs.aws.amazon.com/app-mesh/latest/userguide/envoy-config.html

This change will expose those env variable via helm configuration variable.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #533, opened 2021-11-11
Add support for samplingRate

suniltheta:
**Issue**
https://github.com/aws/aws-app-mesh-roadmap/issues/371

**Description of changes:**
Add back support for samplingRate
This reverts commit 413e594a3f3f85988768e831a4a7bdbf3b409869. #522

This support was originally added in #512


pkg/inject/config.go
```diff
@@ -153,6 +155,8 @@ func (cfg *Config) BindFlags(fs *pflag.FlagSet) {
 		"Enable Envoy X-Ray tracing integration and injects xray-daemon as sidecar")
 	fs.Int32Var(&cfg.XrayDaemonPort, flagXrayDaemonPort, 2000,
 		"X-Ray Agent tracing port")
+	fs.StringVar(&cfg.XraySamplingRate, flagXraySamplingRate, "",
```
cgchinmay, line 158: shouldn't the default value be 0.05 ?

suniltheta: true. Now set the default value 0.05.

pkg/inject/sidecar_builder.go
```diff
@@ -81,6 +84,29 @@ func updateEnvMapForEnvoy(vars EnvoyTemplateVariables, env map[string]string, vn
...
+		samplingRate, ok := env["XRAY_SAMPLING_RATE"]
+		if ok {
+			// `podAnnotations` contains the sampling rate and gets preference over helm configuration
+			// For now delete this value from env so that we can validate before adding again
+			delete(env, "XRAY_SAMPLING_RATE")
+		} else {
+			// `podAnnotations` doesn't contain the sampling rate so get value from helm configuration
+			samplingRate = vars.XraySamplingRate
+		}
+		if samplingRate != "" {
```
cgchinmay, line 98: Do we need this check ? The samplingRate can be empty if the customer has erroneously specified "" value in pod annotations. In that case, we can notify the Customer of this error ?

suniltheta: yes we don't need this check. Now that we implicitly always set this to 0.05 and like you mentioned if the user makes a mistake then we notify via an error.

cgchinmay (approved):
/lgtm

cgchinmay:
@suniltheta What is the behavior with envoy image older than 1.19.1.1 ?

suniltheta:
Envoy image prior to 1.19.1.1, there will be no effect even if we set this value. The X-ray sampling will still pick default value of 0.05.

cgchinmay (approved):
/lgtm

PR #535, opened 2021-11-18
Updated Dockerfile to support multi-arch

cgchinmay:
Updated Makefile to use docker buildx
*Description of changes:*
Added support for building the controller image for muti-arch such as amd64 and arm64


Dockerfile
```diff
@@ -13,7 +15,7 @@ ENV VERSION_PKG=github.com/aws/aws-app-mesh-controller-for-k8s/pkg/version
 RUN GIT_VERSION=$(git describe --tags --dirty --always) && \
     GIT_COMMIT=$(git rev-parse HEAD) && \
     BUILD_DATE=$(date +%Y-%m-%dT%H:%M:%S%z) && \
-    CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build \
```
M00nF1sh, line 16: 1. i think linux should be replaced with TARGETOS
2. do we need below to work?
```
ARG TARGETOS
ARG TARGETARCH
```

cgchinmay: I intentionally didn't add these arguments since on Mac (darwin), TARGETOS could be darwin. Let me try to see what manifest gets built and I will update it

M00nF1sh (approved):
/lgtm

PR #537, opened 2021-11-23
Inject k8s related info to envoy

flashyang:
*Description of changes:*
This change will allow controller to inject k8s platform info to Envoy container via env variable. Basically, we need to inject 3 info for now. 
k8sVersion: we get the k8s version via making a call to k8s server though provided discovery client from clientset
controllerVersion: we already have this info in hand, just inject it
podUid: since controller doesn't know the UID before the pod is running, so we take use of the downward API in k8s and retrieve this info during runtime

### DIFF
--- a/pkg/inject/sidecar_builder.go
+++ b/pkg/inject/sidecar_builder.go
@@ -42,6 +42,8 @@ type EnvoyTemplateVariables struct {
 	StatsDPort               int32
 	StatsDAddress            string
 	StatsDSocketPath         string
+	K8sVersion               string
+	ControllerVersion        string
 }
 
 func updateEnvMapForEnvoy(vars EnvoyTemplateVariables, env map[string]string, vname string) error {
@@ -151,6 +153,9 @@ func updateEnvMapForEnvoy(vars EnvoyTemplateVariables, env map[string]string, vn
 		env["JAEGER_TRACER_PORT"] = vars.JaegerPort
 		env["JAEGER_TRACER_ADDRESS"] = vars.JaegerAddress
 	}
+
+	env["APPMESH_PLATFORM_K8S_VERSION"] = vars.K8sVersion
+	env["APPMESH_PLATFORM_APP_MESH_CONTROLLER_VERSION"] = vars.ControllerVersion
 	return nil
 }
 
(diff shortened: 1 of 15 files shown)


pkg/inject/sidecar_builder.go
```diff
@@ -325,6 +331,19 @@ func envVar(envName, envVal string) corev1.EnvVar {
 	}
 }
 
+// refPodUid is to use the k8s downward API and render pod uid
+func refPodUid(envName string) corev1.EnvVar {
```
cgchinmay, line 335: Could you elaborate little more in comments about the usage ?

cgchinmay (changes requested):
Could you update your branch with latest changes ?

flashyang:
> Could you update your branch with latest changes ?

pkg/inject/sidecar_builder.go
flashyang: Yeah, will add more comments about it

pkg/k8s/utils.go
```diff
@@ -12,3 +17,12 @@ func NamespacedName(obj metav1.Object) types.NamespacedName {
 		Name:      obj.GetName(),
 	}
 }
+
+// ServerVersion returns the version for k8s server
+func ServerVersion(client discovery.ServerVersionInterface) string {
+	versionInfo, err := client.ServerVersion()
+	if err != nil {
+		return unknownK8sVersion
```
cgchinmay, line 25: Would it help to log the error message ? I assume , this will be called only on restart.

cgchinmay (commented):
Could you update the description on why envoy needs this info. Thanks

pkg/k8s/utils.go
flashyang: I would prefer we don't log error here since we are trying out best to get the k8s version here. If we can't get it shouldn't have any side impact on controller itself. So if we log error here might be a lot confusion to user which feels like there is something wrong with controller

cgchinmay: You can set a log level so that it's visible only in Debug mode, if thats the concern.But I am okay with this if its not critical, just a suggestion

flashyang: Thanks for the suggestion. I think we should be good here, if we need we can add it locally :)

cgchinmay (approved):
lgtm

flashyang:
> Could you update the description on why envoy needs this info. Thanks

Update the

PR #549, opened 2021-12-16
Replace busybox image in jaeger init container

sshver:
*Issue #, if available:*

The init container uses a busybox image from docker hub. The private clusters with no outbound internet access are not able to pull this image. 

*Description of changes:*

Replaced the `busybox` image from dockerhub and used the same image from public ECR.

### DIFF
--- a/pkg/inject/jaeger.go
+++ b/pkg/inject/jaeger.go
@@ -39,7 +39,7 @@ const jaegerInitContainerTemplate = `
     "-c",
     "cat <<EOF >> /tmp/envoy/envoyconf.yaml{{ .EnvoyConfig }}EOF\n\ncat /tmp/envoy/envoyconf.yaml\n"
   ],
-  "image": "busybox",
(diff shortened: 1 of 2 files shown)


suniltheta:
Do we still need this busybox? From my understanding we needed jaeger.go to inject an envoy config file containing Zipkin tracer config inside envoy container. Now that AppMesh envoy v1.17+ bootstrap can generate this configuration, I think we can entirely get rid of this jaeger.go & jaeger_tests.go file. 

@cgchinmay what are your thoughts?

cgchinmay:
> Do we still need this busybox? From my understanding we needed jaeger.go to inject an envoy config file containing Zipkin tracer config inside envoy container. Now that AppMesh envoy v1.17+ bootstrap can generate this configuration, I think we can entirely get rid of this jaeger.go & jaeger_tests.go file.
> 
> @cgchinmay what are your thoughts?

@suniltheta In that case we don't need it. If its validated then we can modify the PR to remove the busybox dependency.

PR #558, opened 2022-02-10
Update tests to use ginkgo v2

jeremymill:
*Description of changes:*
Updated import statements in the tests to use ginkgo v2 as v1 is no longer supported. Tested locally with Ginkgo Version 2.1.1


jeremymill:
Oops, turns out the tests were not passing and more changes are needed for migrating v1 to v2.

Issue #556, opened 2022-01-25
Controller fails to propagate `LimitExceededException` from AppMesh


Y0Username:
**Describe the bug**
While creating AppMesh resources, after reaching the resource limit, k8s responds with creation successful while the appmesh controller sits and tries to reconcile the failure to create the resource.

**Steps to reproduce**
1. Create 200 VirtualNodes in a mesh.
2. And then create  201st VirtualNode:
```
 ---
 # should fail
 apiVersion: appmesh.k8s.aws/v1beta2
 kind: VirtualNode
 metadata:
   name: fail
   namespace: howto-k8s-basics
 spec:
   podSelector:
     matchLabels:
       app: colorapp
   listeners:
     - portMapping:
         port: 8080
         protocol: http
   serviceDiscovery:
     dns:
       hostname: fail-headless.howto-k8s-basics.svc.cluster.local
       responseType: ENDPOINTS
```

3. When I execute `kubectl  apply -f ./above_file.yaml`, it responds with `virtualnode.appmesh.k8s.aws/fail created` while in reality the resource creation has failed.

4. AppMesh controller tries to reconcile the resource with following  error: 
```
[...]

PR #559, opened 2022-02-14
Propagate Errors from Reconcile loop

jeremymill:
*Issue #, if available:*
#556 

*Description of changes:*
Errors from the reconcile loop are now recorded as K8s events, which can be viewed with `kubectl describe` or `kubectl get event`. In the example below, `kubectl apply -f make_node.yaml` attempts to create the 201st VirtualNode in a mesh, causing a `LimitExceededException`.

make_node.yaml:
```
kind: VirtualNode
metadata:
    name: limit-fail
    namespace: howto-k8s-http2
spec:
    podSelector:
      matchLabels:
        app: colorapp
    listeners:
      - portMapping:
          port: 8080
          protocol: http
    serviceDiscovery:
      dns:
        hostname: limit-fail-headless.howto-k8s-http2.svc.cluster.local
        responseType: ENDPOINTS
```

The response is still `virtualnode.appmesh.k8s.aws/limit-fail created` so as not to block on resource creation (see https://github.com/aws/aws-app-mesh-controller-for-k8s/issues/556#issuecomment-1021668186).

`kubectl describe VirtualNode/limit-fail -n howto-k8s-http2`

```
Name:         limit-fail
Namespace:    howto-k8s-http2
Labels:       <none>
Annotations:  <none>
API Version:  appmesh.k8s.aws/v1beta2
Kind:         VirtualNode
Metadata:
  Creation Timestamp:  2022-02-14T23:22:16Z
  Finalizers:
    finalizers.appmesh.k8s.aws/aws-appmesh-resources
  Generation:  1
  Managed Fields:
    API Version:  appmesh.k8s.aws/v1beta2
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:finalizers:
          .:
          v:"finalizers.appmesh.k8s.aws/aws-appmesh-resources":
    Manager:      controller
    Operation:    Update
    Time:         2022-02-14T23:22:16Z
    API Version:  appmesh.k8s.aws/v1beta2
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .:
          f:kubectl.kubernetes.io/last-applied-configuration:
      f:spec:
        .:
        f:listeners:
        f:podSelector:
          .:
          f:matchLabels:
            .:
            f:app:
        f:serviceDiscovery:
          .:
          f:dns:
            .:
            f:hostname:
            f:responseType:
    Manager:         kubectl-client-side-apply
    Operation:       Update
    Time:            2022-02-14T23:22:16Z
  Resource Version:  978195
  UID:               51ac0306-90e9-4073-94f9-9cb52ac5050a
Spec:
  Aws Name:  limit-fail_howto-k8s-http2
  Listeners:
    Port Mapping:
      Port:      8080
      Protocol:  http
  Mesh Ref:
    Name:  howto-k8s-http2
    UID:   f611cc36-8ce4-4c6a-b891-1b5f17bd980a
  Pod Selector:
    Match Labels:
      App:  colorapp
  Service Discovery:
    Dns:
      Hostname:       limit-fail-headless.howto-k8s-http2.svc.cluster.local
      Response Type:  ENDPOINTS
Events:
  Type     Reason          Age                From         Message
  ----     ------          ----               ----         -------
  Warning  ReconcileError  9s (x12 over 19s)  VirtualNode  LimitExceededException: You've reached the limit on number of VirtualNodes(200) for the mesh howto-k8s-http2.
```


controllers/appmesh/virtualservice_controller.go
```diff
@@ -75,6 +78,7 @@ func (r *virtualServiceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	}); err != nil {
 		return err
 	}
+	r.recorder = mgr.GetEventRecorderFor("VirtualService")
```
cgchinmay, line 81: This r.recorder field is part of the virtualServiceReconciler and we initiate all the fields when we get reference to this virtualServiceReconciler, like here
```
vsReconciler := appmeshcontroller.NewVirtualServiceReconciler(mgr.GetClient(), finalizerManager, referencesIndexer, vsResManager, ctrl.Log.WithName("controllers").WithName("VirtualService"))

func NewVirtualServiceReconciler(k8sClient client.Client, finalizerManager k8s.FinalizerManager, referencesIndexer references.ObjectReferenceIndexer, vsResManager virtualservice.ResourceManager, log logr.Logger, recorder record.EventRecorder) *virtualServiceReconciler {
	return &virtualServiceReconciler{
		k8sClient:                             k8sClient,
		finalizerManager:                      finalizerManager,
		referencesIndexer:                     referencesIndexer,
		vsResManager:                          vsResManager,
		enqueueRequestsForMeshEvents:          virtualservice.NewEnqueueRequestsForMeshEvents(k8sClient, log),
		enqueueRequestsForVirtualNodeEvents:   virtualservice.NewEnqueueRequestsForVirtualNodeEvents(referencesIndexer, log),
		enqueueRequestsForVirtualRouterEvents: virtualservice.NewEnqueueRequestsForVirtualRouterEvents(referencesIndexer, log),
		log:                                   log,
	}
}

Similarly we could also pass in recorder to this function
vsReconciler := appmeshcontroller.NewVirtualServiceReconciler(mgr.GetClient(), finalizerManager, referencesIndexer, vsResManager, ctrl.Log.WithName("controllers").WithName("VirtualService"), mgr.GetEventRecorderFor("VirtualService"))
```
Would like to know your thoughts, just to have a consistent pattern, suggesting this approach. Will be easier to track.

cgchinmay (commented):
Thanks for the PR ! Just few comments

controllers/appmesh/virtualservice_controller.go
jeremymill: Agreed, much easier to track. Will revise soon!

controllers/appmesh/cloudmap_controller.go
```diff
@@ -83,7 +89,11 @@ func (r *cloudMapReconciler) reconcile(ctx context.Context, req ctrl.Request) er
 	if !vNode.DeletionTimestamp.IsZero() {
 		return r.cleanupCloudMapResources(ctx, vNode)
 	}
-	return r.reconcileVirtualNodeWithCloudMap(ctx, vNode)
+	if err := r.reconcileVirtualNodeWithCloudMap(ctx, vNode); err != nil {
+		r.recorder.Event(vNode, corev1.EventTypeWarning, "ReconcileError", err.Error())
```
ar4096, line 93: This one is pretty important. We have a bug related to this. Do you know if the inner function (reconcileVirtualNodeWithCloudmap) or the caller log the details of the failure ?

controllers/appmesh/cloudmap_controller_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/record"
+	testclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"testing"
+)
+
+func Test_cloudMapReconciler_reconcile(t *testing.T) {
```
ar4096, line 21: Any thoughts on how we can potentially share some across tests and modules so that LOC per test are less ?

ar4096 (approved):

controllers/appmesh/gatewayroute_controller.go
```diff
@@ -83,7 +89,11 @@ func (r *gatewayRouteReconciler) reconcile(ctx context.Context, req ctrl.Request
 	if !gr.DeletionTimestamp.IsZero() {
 		return r.cleanupGatewayRoute(ctx, gr)
```
Y0Username, line 90: We should probably come back and add events here too.

controllers/appmesh/gatewayroute_controller.go
```diff
@@ -83,7 +89,11 @@ func (r *gatewayRouteReconciler) reconcile(ctx context.Context, req ctrl.Request
...
 	}
-	return r.reconcileGatewayRoute(ctx, gr)
+	if err := r.reconcileGatewayRoute(ctx, gr); err != nil {
+		r.recorder.Event(gr, corev1.EventTypeWarning, "ReconcileError", err.Error())
```
Y0Username, line 93: I'm not convinced that this should be `WARN` level, failing to create any resource will technically impact the routing in the cluster. Thoughts @jeremymill and @cgchinmay 

Also, could we rename `ReconcileError`, can we get the error type from the `err` Object? Could it be specific such as`BadRequestException` or `LimitExceededException` or whatever else?

jeremymill: Since we only have Warning events and Normal events, I'm not sure if Normal is more appropriate than Warning here.

I could make a utility function to extract the error type (`err.split(":")[0]`, effectively) but this probably won't behave well unless all errors are formatted this way.

controllers/appmesh/gatewayroute_controller.go
jeremymill: Good point, if deletion / cleanup fails we probably want events for that.

controllers/appmesh/gatewayroute_controller.go
cgchinmay: @jeremymill EventType is a string so you can pass "Error" string instead or have our own custom ErrorType to distinguish between Errors and warnings. Will be helpful in future to decide based on the error type.

controllers/appmesh/cloudmap_controller.go
ar4096: @cgchinmay If the reconcile call fails, where does the retry happen ? I ask because do we want to guard against an Event storm if retries are done at a higher level.

cgchinmay: From what I know, the controller-runtime invokes the Reconcile loop every time we return an error or for any of the create, update, delete events based on how we have configured the watcher in SetupWithManager. It does have a built-in back off logic. So if we want to stop the reconcile loop then I guess we need to return nil. We can decide this based on the error type whether it's intermittent or not, since doing it based on number of retries could be hard. Will have to test this logic.

controllers/appmesh/gatewayroute_controller.go
jeremymill: I tried passing "Error" string instead of EventType, this causes the event to not appear.

cgchinmay: Okay, will have to check where is this validated since it was a string type. But in that case, I guess we will have to use only the built-in EventTypes for now

controllers/appmesh/gatewayroute_controller.go
Y0Username: Not a blocker, we can follow up on this.

controllers/appmesh/gatewayroute_controller.go
Y0Username: That's lame that `EventType` doesn't support other types, let's keep it `WARN` for now and come back and fix it.

What members/methods does the `err` Object have? I understand that `err.Error()` is a string, but does it also have `err.Type()`? If not, let's leave it at `ReconcileError`, substrings are hacky.

ar4096:
Yay!

From: jeremymill ***@***.***>
Reply-To: aws/aws-app-mesh-controller-for-k8s ***@***.***>
Date: Wednesday, May 11, 2022 at 12:39 PM
To: aws/aws-app-mesh-controller-for-k8s ***@***.***>
Cc: "Rana, Ahad" ***@***.***>, Review requested ***@***.***>
Subject: Re: [aws/aws-app-mesh-controller-for-k8s] Propagate Errors from Reconcile loop (PR #559)


Merged #559<https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/559> into master.

PR #561, opened 2022-02-24
IPv6 crd changes

sshver:
*Issue #, if available:*

*Description of changes:*
IPv6 CRD changes


main.go
```diff
@@ -175,6 +176,28 @@ func main() {
 		os.Exit(1)
 	}
 
+	kubeConfigArgs := kubeConfig.ExecProvider.Args
```
sshver, line 179: This is a very hacky approach to get the cluster-name from KubeConfig. I could not find any other ways to get the clusterName. None of the APIs were in here: https://pkg.go.dev/k8s.io/client-go outputs the cluster name.

sshver:
The unit tests are failing as the AppMesh SDK models are not updated in preview.

apis/appmesh/v1beta2/mesh_types.go
```diff
@@ -81,6 +81,14 @@ type MeshSpec struct {
 	// Required if the account ID is not your own.
 	// +optional
 	MeshOwner *string `json:"meshOwner,omitempty"`
+	// +optional
+	MeshServiceDiscovery *MeshServiceDiscovery `json:"meshServiceDiscovery,omitempty"`
```
cgchinmay, line 85: The field names should align with appmesh sdk wherever possible.

apis/appmesh/v1beta2/virtualnode_types.go
```diff
@@ -335,6 +338,9 @@ type DNSServiceDiscovery struct {
 	// +kubebuilder:validation:Enum=ENDPOINTS;LOADBALANCER
 	// +optional
 	ResponseType *string `json:"responseType,omitempty"`
+	// The ipPreference for the virtual node.
+	// +optional
+	IpPreference *string `json:"ipPreference,omitempty"`
```
cgchinmay, line 343: Add a kubebuilder annotation for allowed values for this field, the way we have for ResponseType

main.go
cgchinmay: We are using this as Fallback approch right to determine IpFamily. Customers can always specify using yaml spec ?

sshver: Yes.

apis/appmesh/v1beta2/mesh_types.go
sshver: AppMesh SDK is using `MeshServiceDiscovery` in the field.

main.go
ar4096: This is quite brittle no ? Do we need a fallback ? Is this because of existing deployments in the field ?

apis/appmesh/v1beta2/virtualnode_types.go
ar4096: Good call @cgchinmay

main.go
sshver: EKS currently doesn't support dual stack. The IpPreference values like `IPv4_PREFERRED` and `IPv6_PREFERRED` (which are a mix of IPv4 and IPv6) can't be supported as a EKS cluster could either be IPv4 or IPv6. We decided to expose only IPv4 or IPv6 options to the customers. 

Given IpPreference can either be - IPV4_only, IPV6_only and option to set IpPreference is optional. By default, we decided to derive ipFamily the option from underlying EKS cluster. If ipFamily IPV6, set IpPreference to IPV6_only. If ipFamily IPV4, set IpPreference to IPV4_only.

When the option is not set App Mesh uses a mix of IPv4_PREFFERED, IPv6_PREFFERED, and IPv4_ONLY. Therefore we need a default value according to the ipFamily used by the cluster.

apis/appmesh/v1beta2/virtualnode_types.go
sshver: fixed in new revision.

main.go
ar4096: I guess the question is why is it optional ? If you want to stick with this decision, can you think of a better option that introspecting the command line switch ?

Issue #560, opened 2022-02-21
Unable to update cloudmap DNS TTL


mindbergh:
**Describe the bug**
There are two issues regarding update cloudmap DNS ttl.

1. [helm expects float64](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/config/helm/appmesh-controller/templates/deployment.yaml#L77) whereas the [appmesh controller expects int64](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/cloudmap/config.go#L17):

For example, if I give a TTL of 10 in the helm release, it will be ignored by the template.

2. Even if the appmesh controller's deployment get updated to include the TTL argument, the reconcile will not update the existing cloudmap service. Looks like the [reconciler logic](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/cloudmap/resource_manager.go#L104) only takes ttl into consideration when creating a new cloudmap service but doesn't update if there is a change.

I had to manually delete the cloudmap service and let the controller reconcile to recreate one in order to apply the DNS TTL change.

**Steps to reproduce**
1. Using the helm to update the cloudmap DNS TTL.
2. The controller deployment will not be updated.
3. The DNS TTL in cloudmap service will not be updated.

**Expected outcome**
1. The helm release is able to update the cloudmap TTL of the appmesh controller deployment.
2. Appmesh controller is able to update existing cloudmap service's DNS ttl.

**Environment**

* App Mesh controller version: helm chart v1.4.4, controller: v1.4.2
* Kubernetes version: 1.21
* Using EKS (yes/no), if so version? yes

**Additional Context**:
N/A

melaniesifen:
This issue for setting the DNS TTL has been resolved with the latest eks appmesh-controller helm chart release. This field will not update existing CloudMap services. Existing services can be updated using the [AWS CloudMap API](https://docs.aws.amazon.com/cloud-map/latest/api/API_UpdateService.html)

PR #574, opened 2022-04-07
Cloudmapdnsttl

melaniesifen:
*Issue #, if available:*
[560](https://github.com/aws/aws-app-mesh-controller-for-k8s/issues/560)
*Description of changes:*

- Updated helm chart mismatched field
- created reconcile logic to update cloudmap dns ttl setting
- added util functions for cloudmap, should be used to avoid time.sleep


pkg/cloudmap/utils.go
```diff
@@ -24,3 +37,61 @@ func ShouldPodBeInEndpoints(pod *corev1.Pod) bool {
...
+				status := f()
+				if status == SUCCESS {
+					operationStatus <- status
+					return
+				}
+			}
+		}
+	}()
+	return <-operationStatus
+}
```
paulyeo21, line 61: i dont think your function will timeout appropriately if the function passed in `f()` exceeds the timeout duration. for example if you run this test which is a modification of what you have:

```golang
func TestAwaitOperationSuccess(t *testing.T) {
	tests := []struct {
		name    string
		timeout time.Duration
		tick    time.Duration
		want    string
	}{
		{
			name:    "Await function should succeed",
			timeout: 10 * time.Second,
			tick:    1 * time.Second,
			want:    SUCCESS,
		},
		{
			name:    "Await function should timeout",
			timeout: 2 * time.Second,
			tick:    1 * time.Second,
			want:    TIMEOUT,
		},
	}
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			status := AwaitOperationSuccess(tt.timeout, tt.tick, func() string {
				time.Sleep(3 * time.Second)
				return SUCCESS
			})

			assert.Equal(t, tt.want, status)
		})
	}
}
```
output:
```
=== RUN   TestAwaitOperationSuccess
=== RUN   TestAwaitOperationSuccess/Await_function_should_succeed
=== RUN   TestAwaitOperationSuccess/Await_function_should_timeout
    utils_test.go:37: 
        	Error Trace:	utils_test.go:37
        	Error:      	Not equal: 
        	            	expected: "TIMEOUT"
        	            	actual  : "SUCCESS"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-TIMEOUT
        	            	+SUCCESS
        	Test:       	TestAwaitOperationSuccess/Await_function_should_timeout
--- FAIL: TestAwaitOperationSuccess (8.00s)
    --- PASS: TestAwaitOperationSuccess/Await_function_should_succeed (4.00s)
    --- FAIL: TestAwaitOperationSuccess/Await_function_should_timeout (4.00s)
FAIL
FAIL	command-line-arguments	8.670s
FAIL
```

indicating that the timeout of 2 seconds is not respected bc the function passed in sleeps for 3 seconds. there are a couple of ways to do this, but i believe what you want is to create two separate goroutines that are competing for completion and receive/return the faster of the two. something like the following:

```golang
func AwaitOperationSuccess(timeout time.Duration, tick time.Duration, f func() string) string {
	res := make(chan string, 1)

	go func() {
		time.Sleep(timeout) // or <-time.After(timeout) doesnt really matter
		res <- TIMEOUT
	}()

	go func() {
		for _ = range time.Tick(tick) {
			status := f()
			if status == SUCCESS {
				res <- SUCCESS
                                 return
			}
		}
	}()

	return <-res
}
```

paulyeo21: what is the expectation for the passed-in function? its a bit odd that we are creating custom output strings in the utils file that vary from the usages in say `resource_manager.go` but we have a line ```status == SUCCESS``` where we expect the output of the passed-in function to be a certain value. i see from some digging that the output of the usage in `resource_manager.go` can be these enums https://github.com/aws/aws-sdk-go/blob/main/service/servicediscovery/api.go#L6637

controllers/appmesh/cloudmap_controller.go
```diff
@@ -70,7 +70,7 @@ func (r *cloudMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Named("cloudMap").
 		For(&appmesh.VirtualNode{}).
 		Watches(&k8s.NotificationChannel{Source: r.podEventNotificationChan}, r.enqueueRequestsForPodEvents).
-		WithOptions(controller.Options{MaxConcurrentReconciles: 3}).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 1}).
```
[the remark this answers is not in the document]
melaniesifen: I changed this because I was receiving reconcile errors for duplicate update requests for the cloudmap services.

pkg/cloudmap/utils.go
melaniesifen: For the strings (SUCCESS, TIMEOUT, FAIL), I created them in the utils.go file so that I could use them in the resource_manager.go, utils.go, and (eventually) the related integration tests. I was thinking I could use those enums, but there isn't a TIMEOUT one that would indicate that the related async operation will possibly succeed but is taking too long. I also thought it looked better to define those as constants since they would be used frequently.

pkg/cloudmap/resource_manager.go
```diff
@@ -288,6 +296,49 @@ func (m *defaultResourceManager) createCloudMapService(ctx context.Context, vn *
...
+		m.log.V(1).Info("Sending cloudMap service update request")
+		operationId, err := m.updateCloudMapServiceUnderPrivateDNSNamespace(ctx, vn, nsSummary, serviceName, cloudMapTTL, svcSummary)
+		if err != nil {
+			return nil, err
+		}
+		// wait for update to succeed or timeout and udpate on next reconcile loop
+		operationInput := &servicediscovery.GetOperationInput{
+			OperationId: operationId,
+		}
+		status := AwaitOperationSuccess(30, 1, func() string {
```
ar4096, line 327: I don't understand why need this logic at all. The aws sdk for go has retry and timeout options at the client level. We should be configuring these, not adding our own logic for this. (see https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/retries-timeouts/)

controllers/appmesh/cloudmap_controller.go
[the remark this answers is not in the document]
ar4096: It would be good to understand the root cause of these errors. Increasing concurrency to > 1 means concurrent reconciliation runs but not concurrent reconciliation of the same CRD. So, is there something in our model that is causing two CRD updates to converge to one cloudmap resource ?

pkg/cloudmap/utils.go
```diff
@@ -24,3 +37,69 @@ func ShouldPodBeInEndpoints(pod *corev1.Pod) bool {
 		return true
 	}
 }
+
+func AwaitOperationSuccess(timeoutSeconds time.Duration, tickerSeconds time.Duration, f func() string) string {
+	operationStatus := make(chan string, 1)
```
ar4096, line 42: This should just be something like 

endTime = wallClock + timeoutSeconds

while (wallClock < endTime) { 
   checkStatus();
   if (state == GOOD || FAIL) { 
      return;
   }
   return TIMEOUT;
}

ar4096, line 42: This should just be something like 

endTime = wallClock + timeoutSeconds

while (wallClock < endTime) { 
   checkStatus();
   if (state == GOOD || FAIL) { 
      return;
   }
   return TIMEOUT;
}

ar4096: Also make sure you control Client timeout. In other words, client shouldn't go out and use a 30 second timeout for every attempt.

ar4096: Also make sure you control Client timeout. In other words, client shouldn't go out and use a 30 second timeout for every attempt.

pkg/cloudmap/resource_manager.go
ar4096: Just to clarify after my conversation with @melaniesifen. My question centered around why we need AwaitOperationSuccess when the SDK provides retries and timeouts. It turns out that Cloudmap has an async API: You do an operation, you get a request id back and you poll (via a different API) for the status of that request. Hence why @melaniesifen added that function. Which now makes sense. My recommendation to her is that this is a special purpose function specifically designed to poll for the Cloudmap status change (from PENDING to SUCCESS or FAIL). So, might want to either move the call to cloudMapSDK.GetOperation inside the function or change the prototype of the lambda to take the appropriate enumerated return type. Also, we might want to change the name of the function to better reflect its purpose... Something like PollCloudmapStatus ... etc. Lastly, the interior of the function doesn't need to have a go routine since the call to Await... is blocking. i.e. you are sitting in a loop polling for status until you are out of time. The previous implementation (with 2 go routines, one for timeout and one for the actual blocking REST call) was tricky because you could timeout and exit the main function but the poll routine might keep running or be blocked. Generally, we don't want to leave orphaned and blocked go routines or threads just hanging around if we can help it. Which brings me to my last point: lets still understand timeouts and retries for the HTTP Client. If the HTTP client has a default timeout of 30 seconds and a retry count of 3 etc., we might be out to lunch for 60 seconds when we called AwaitOperationSuccess with an expected timeout of 30 seconds (for example). I don't we can completely avoid something like this, but we should reduce the timeout overshoot if we can help it.

Issue #567, opened 2022-03-30
Configure runAsUser=0 on injected proxyinit container security context


rimaulana:
**Is your feature request related to a problem?**
When pod has podSecurityContext that define the default runAsUser other than 0, proxyinit initContainers are failing with the following error message 

```
Fatal: can't open lock file /run/xtables.lock: Permission denied
```

**Describe the solution you'd like**
It is understandable that proxyinit will execute iptables command to deploy some traffic redirection rules and it requires to run as root user. If this is a hard requirement, I would want the appmesh-controller to inject this as part of container security context. The current injected container security context is

```
securityContext:
  capabilities:
    add:
    - NET_ADMIN
```

What I would like to see is to look like

```
securityContext:
  runAsUser: 0
  capabilities:
    add:
    - NET_ADMIN
```

petrzjunior:
Despite the pull request years ago, this is still an open issue.
I am using AWS AppMesh in a restricted environment with security context set to non-root user. This is currently blocking me, I need to set `securityContext` to root for the whole pod where a container-specific `securityContext` for Envoy would be sufficient.
I basically do not a scenario where NOT setting the container `securityContext` to root would work. It is trying to lock `/run/xtables.lock` and fails.
I am not proficient in Go to open a PR, perhaps some of the maintainers could help us? @ysdongAmazon @srinivas-kini

PR #576, opened 2022-04-13
Resolving ticket #567, set default runAsUser=0 to init proxy container

EdwardXF:
*Issue #, if available:*
When pod has podSecurityContext that define the default runAsUser other than 0, proxyinit initContainers are failing with the following error message

`Fatal: can't open lock file /run/xtables.lock: Permission denied`

*Description of changes:*
Added default runAsUser value in proxy.go and loaded it into init proxy container's security context in init_proxy.go. Changed the unit tests accordingly.

### DIFF
--- a/pkg/inject/config.go
+++ b/pkg/inject/config.go
@@ -164,7 +164,7 @@ func (cfg *Config) BindFlags(fs *pflag.FlagSet) {
 		"Number of seconds after Envoy has started before readiness probes are initiated")
 	fs.Int32Var(&cfg.ReadinessProbePeriod, flagReadinessProbePeriod, 10,
 		"How often (in seconds) to perform the readiness probe on Envoy container")
-	fs.StringVar(&cfg.InitImage, flagInitImage, "840364872350.dkr.ecr.us-west-2.amazonaws.com/aws-appmesh-proxy-route-manager:v6-prod",
+	fs.StringVar(&cfg.InitImage, flagInitImage, "840364872350.dkr.ecr.us-west-2.amazonaws.com/aws-appmesh-proxy-route-manager:v7-prod",
 		"Init container image.")
 	fs.StringVar(&cfg.IgnoredIPs, flagIgnoredIPs, "169.254.169.254",
 		"Init container ignored IPs.")


pkg/inject/proxy.go
```diff
@@ -8,6 +8,7 @@ import (
 )
 
 const (
+	defaultRunAsUser          = 0
```
Y0Username, line 11: Why does proxy need to run as root?

pkg/inject/init_proxy.go
```diff
@@ -51,6 +51,7 @@ const proxyInitContainerTemplate = `
 `
 
 type InitContainerTemplateVariables struct {
+	RunAsUser          int64
```
Y0Username, line 54: add this value in the `securityContext` above.

pkg/inject/proxy.go
```diff
@@ -43,6 +44,8 @@ func (m *proxyMutator) mutate(pod *corev1.Pod) error {
 }
 
 type proxyConfig struct {
+	// RunAsUser for init container
+	runAsUser int64
```
Y0Username, line 48: Why is the init container config under proxy configs?

Y0Username (changes requested):

pkg/inject/init_proxy.go
EdwardXF: The controller will take the template and convert it to a JSON string, no matter what type declared here will become string when calling json.unmarshal function and in K8s corev1 container runAsUser field does not take String as an input, so I have to add this directly to config instead of using template.

pkg/inject/proxy.go
EdwardXF: Although this is defined in proxy.go, the proxy config is only used in initProxyMutator and cniProxyMutator, and in cniProxyMutator, this field is not used so runAsUser is not set there.

pkg/inject/proxy.go
EdwardXF: I am not sure why it is there, but I only find this variable used in initProxyMutator and cniProxyMutator so it should not affect the envoy config.

PR #594, opened 2022-05-13
Update default to app version 1.4.3 (was incorrectly 1.4.2)

jeremymill:
*Issue #, if available:*
App version regressed to 1.4.2.

*Description of changes:*
Fixes image version back to 1.4.3.


paulyeo21 (approved):
chatted with @jeremymill offline about this fix, looks good to me!

Issue #489, opened 2021-07-13
Feature to make sure the envoy container always start and work before the application container launch.


yws-ss:
**Is your feature request related to a problem?**

According to Kubernetes logical,  all containers in a Pod are being started in parallel and there is no way to define that one container must be started after other container (excluded the initContainer). Dig into the [code](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/kuberuntime_manager.go#L924), seems kubernetes launch the containers by order:

```
for _, idx := range podContainerChanges.ContainersToStart {
		start("container", metrics.Container, containerStartSpec(&pod.Spec.Containers[idx]))
	}
```
Currently, the envoy container is run as sidecar container with the application container in the pod, and is injected at the last order of Pod spec automatically . Base on  kubernetes logical, it have some chance that application container start before the proxy container activate. Which cause the application container might have outgoing network issue before the envoy pod start to work.

**Describe the solution you'd like**

From Istio [changelog](https://istio.io/latest/news/releases/1.7.x/announcing-1.7/change-notes/), they roll out a feature that allow user to add and option `holdApplicationUntilProxyStarts`, which causes the sidecar injector to inject the sidecar at the start of the pod’s container list and configures it to block the start of all other containers until the proxy is ready. 

Maybe, we could consider to add the similar function in the AWS app mesh.

**Describe alternatives you've considered**

I can totally understand this might be a very race scenario, moreover, this can be avoid by adding retry mechanism in application logical. However, we still want some feature that is simple to use or configure.

cgchinmay:
@yws-ss Will take a look and get back to you with an update

fbuinosquy1985:
I need this , the Vault Agent container injector fails in combination with the Aws Mesh envoy even if i set on annotations [vault.hashicorp.com/agent-init-first](https://www.vaultproject.io/docs/platform/k8s/injector/annotations#vault-hashicorp-com-agent-init-first)

cgchinmay:
Adding the right folks @jeremymill @Y0Username

paulyeo21:
i realize this post and some of these comments are from a while ago, apologies for the wait for those users that have been waiting for this capability. it has been released on v1.6.0 pls try it by enabling the feature flag:

`--set sidecar.waitUntilProxyReady=true`

PR #598, opened 2022-05-26
waitUntilProxyReady feature

paulyeo21:
*Issue #, if available:*

https://github.com/aws/aws-app-mesh-controller-for-k8s/issues/489

*Description of changes:*

Enable pod postStart hook to delay application startup until proxy is ready to accept traffic

Gated with a feature flag:
1) ensure envoy container is first in pod lifecycle
2) add a postStart hook that polls envoy server is ready

Integration spec:
1) new suite to validate behavior
2) includes docker images for frontend and backend apps used in test

Testing instructions:

1) pull branch locally
2) build controller image
```
TARGETPLATFORM=linux/amd64 docker build -t amazon/appmesh-controller .
```
3) tag controller image version
```
docker tag amazon/appmesh-controller:latest $AWS_ACCOUNT_ID.dkr.ecr.$AWS_REGION.amazonaws.com/amazon/appmesh-controller:sidecar-init-v1.0.10
```
4) push to ecr repository
```
5) docker push $AWS_ACCOUNT_ID.dkr.ecr.$AWS_REGION.amazonaws.com/amazon/appmesh-controller:sidecar-init-v1.0.10
```
6) from repo root install
```
helm upgrade appmesh-controller ./config/helm/appmesh-controller \
    --namespace appmesh-system \
    --set region=$AWS_REGION \
    --set serviceAccount.create=false \
    --set serviceAccount.name=appmesh-controller \
    --set image.repository=$AWS_ACCOUNT_ID.dkr.ecr.$AWS_REGION.amazonaws.com/amazon/appmesh-controller
    --set image.tag=sidecar-init-v1.0.10
    --set sidecar.waitUntilProxyReady=true
```
7) run the integration test:
```
ginkgo -v -r test/integration/sidecar -- --cluster-kubeconfig=${dir} --cluster-name=${name} --aws-region=$AWS_REGION --aws-vpc-id=${vpc}
```


test/integration/sidecar/sidecar_stack.go
```diff
@@ -0,0 +1,448 @@
...
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+const (
+	defaultFrontendImage = "public.ecr.aws/b7m0w2t6/color-fe-app:2.0.3"
+	defaultBackendImage  = "public.ecr.aws/b7m0w2t6/color-be-app:2.0.2"
```
paulyeo21, line 25: these are the application images in the `/test_app/sidecar-backend` and `/test_app/sidecar-frontend`

test/integration/sidecar/sidecar_stack.go
```diff
@@ -0,0 +1,448 @@
...
+		role := &rbacv1.Role{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "default",
+				Namespace: s.testName,
+			},
+			Rules: []rbacv1.PolicyRule{
+				{
+					APIGroups: []string{""},
+					Resources: []string{"pods"},
+					Verbs:     []string{"list", "update"},
```
paulyeo21, line 109: we need this role and the subsequent role binding to get a list of the pods and update the pod annotation from within the app container

test/integration/sidecar/sidecar_stack.go
```diff
@@ -0,0 +1,448 @@
...
+		f.Logger.Error("failed to delete namespace")
+	}
+
+	if err := f.K8sClient.Delete(ctx, s.mesh); err != nil {
+		f.Logger.Error("failed to delete mesh")
+	}
+}
+
+func (s *SidecarStack) pollPodUntilCondition(ctx context.Context, podName string, condition corev1.PodPhase) error {
+	return wait.Poll(1*time.Second, 180*time.Second, func() (done bool, err error) {
```
paulyeo21, line 426: this is an arbitrary timeout

test/integration/sidecar/sidecar_stack.go
```diff
@@ -0,0 +1,448 @@
...
+	})
+}
+
+func newStringPtr(s string) *string {
+	return &s
+}
+
+func newBoolPtr(b bool) *bool {
+	return &b
+}
```
paulyeo21, line 448: these exist bc theres not a pretty way to do it inline

test/integration/sidecar/sidecar_stack_test.go
```diff
@@ -0,0 +1,50 @@
...
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		AfterEach(func() {
+			stack.cleanup(ctx, f)
+		})
+
+		It("should have the color annotation", func() {
+			stack.createMeshAndNamespace(ctx, f)
+			stack.createBackendResources(ctx, f)
```
paulyeo21, line 36: im waiting until backend pods are bound to a node to ensure that the test does not fail bc the backend is not ready. then the frontend is provisioned and immediately tested for the annotation after its pods are bound to a node

test/integration/test_app/sidecar-frontend/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    rm -rf /var/cache/yum
+
+RUN curl -LO https://golang.org/dl/go1.17.1.linux-amd64.tar.gz && \
+    tar -C /usr/local -xzvf go1.17.1.linux-amd64.tar.gz
+
+ENV PATH="${PATH}:/usr/local/go/bin"
+ENV GOPATH="${HOME}/go"
+ENV PATH="${PATH}:${GOPATH}/bin"
+
+ENV GOPROXY=direct
```
sshver, line 14: I have observed removing `ENV GOPROXY=direct` speeds up the build process. But I think its a small app and would not make much of a difference.

config/helm/appmesh-controller/values.yaml
```diff
@@ -39,6 +39,7 @@ sidecar:
     # sidecar.probes: Envoy Readiness Probe
     readinessProbeInitialDelay: 1
     readinessProbePeriod: 10
+  waitUntilProxyStarts: false
```
kiranmeduri, line 42: why isn't the default `true`?

config/helm/appmesh-controller/templates/deployment.yaml
```diff
@@ -113,6 +113,7 @@ spec:
         - --sidecar-log-level={{ .Values.sidecar.logLevel }}
         # this must be same as livenessProbe port which can be configured 
         - --health-probe-port={{ .Values.livenessProbe.httpGet.port }}
+        - --wait-until-proxy-starts={{ .Values.sidecar.waitUntilProxyStarts }}
```
kiranmeduri, line 116: is it until-start or until-ready? Do we wait for Envoy to be bootstrapped with config from App Mesh XDS?

pkg/inject/envoy.go
```diff
@@ -107,7 +108,13 @@ func (m *envoyMutator) mutate(pod *corev1.Pod) error {
 	if m.mutatorConfig.enableSDS && !isSDSDisabled(pod) {
 		mutateSDSMounts(pod, &container, m.mutatorConfig.sdsUdsPath)
 	}
-	pod.Spec.Containers = append(pod.Spec.Containers, container)
+
+	if m.mutatorConfig.waitUntilProxyStarts {
+		pod.Spec.Containers = append([]corev1.Container{container}, pod.Spec.Containers...)
```
kiranmeduri, line 113: Is there any reason we should not always add Envoy to the front of the list? Can we get rid of this conditional and reduce testing scope?

pkg/inject/sidecar_builder.go
```diff
@@ -190,6 +191,14 @@ func buildEnvoySidecar(vars EnvoyTemplateVariables, env map[string]string) (core
 		},
 	}
 
+	if vars.WaitUntilProxyStarts {
+		envoy.Lifecycle.PostStart = &corev1.Handler{
+			Exec: &corev1.ExecAction{Command: []string{
+				"sh", "-c", fmt.Sprintf("until curl -s http://localhost:%d/server_info | grep state | grep -q LIVE; do sleep 1; done", vars.AdminAccessPort),
```
kiranmeduri, line 197: Is there no simple CLI command in [agent](https://aws.amazon.com/about-aws/whats-new/2022/02/aws-app-mesh-agent-envoy/) to do this? We should remove curl from the image and then this will break. @abaptiste can you guide us here.

kiranmeduri (dismissed):

config/helm/appmesh-controller/templates/deployment.yaml
paulyeo21: yeah i can see until-ready to be a more apt name. yes it is supposed to wait until the sidecars have the mesh configs

pkg/inject/sidecar_builder.go
paulyeo21: curl is something we already use in our readiness probe https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/inject/sidecar_builder.go#L224, however, i agree with your point, because im actually running into an issue where sleep is not found in the envoy container (maybe because of path issues, something im still investigating). if there are alternative executables or agent commands you referenced that are better, pls let me know. is there more information on agent? I followed that link but it doesnt say much

pkg/inject/envoy.go
paulyeo21: the only thing i can think of is that some kubectl commands like kubectl exec default to the first app container and customers may find it strange to see the order changed due to the injection. does anyone else know other implications for changing the order @ar4096 @bendu @Y0Username @joesbigidea ? that being said, it may be safer to tie this behavior to the feature flag in the situation that there are other implications

test/integration/test_app/sidecar-frontend/Dockerfile
paulyeo21: good to know-- ill play around with it, thanks

pkg/inject/sidecar_builder.go
ar4096: To add to @kiranmeduri 's point, the concern is the vendor upstream of us (agent) controls the container image we are leveraging, so they could break us unintentionally or without us knowing. Even if the agent team vends an _isReady_  (example) script for us to invoke (which in turn calls curl), at least there is a formal contract between us that they cannot break. Or we need to be able to enforce the requirement that _curl_ be present in the image somehow. I don't know if there is a way to do the latter.  _sleep_ is a related but different issue IMHO. I guess they could ship with no shell at all, but I think sleep is _ususally_ available in busybox or even the most chroot'd jail ?

config/helm/appmesh-controller/templates/deployment.yaml
ar4096: I second that. we should probably change the parameter name to be more consistent with the condition we are looking for.

pkg/inject/sidecar_builder.go
joesbigidea: I think the curl concern is reasonable but given that dependency already exists in the controller It seems to me it shouldn't hold up this change. We should maybe spin up a separate ticket with the agent folks to get these scripts encapsulated as @ar4096 suggests.

pkg/inject/envoy.go
ar4096: We should think on this some more @paulyeo21. Remember networking is already broken because the init container already redirected traffic to our sidecar, so whether _waitUntilReady_ is enabled or not, we still might want to be the first container out of the gate.

pkg/inject/sidecar_builder.go
abaptiste: Sorry I'm late to the PR.  The `curl` binary in the image is our replacement of the standard tool.  It's a static golang binary that replaces the traditional binary to eliminate quite a few system (34) libraries.

Secondly, the binaries present in the image do not include sleep or timeout.

I would recommend leveraging the agent itself to perform the check since it's responsible for starting Envoy and already does health checks of the proxy.  You could build a equivalent behavior of `until curl -s ...` into the supervisor  binary itself and define some IPC between the process running Envoy and the command line invocation to get the proxy's state.

paulyeo21: hi @abaptiste, thx for the comment. what is the supervisor binary here you are referring to? also i looked at the agent codebase and im a bit fuzzy on what you are recommending in regards to the command line invocation. do you mind elaborating where that would live? also if we do go down this path, as @kiranmeduri mentioned here in the quip design comment https://quip-amazon.com/5JRzAMebdaEO#IOK9DA9Cpzk, this would couple the controller release with the envoy image release. what is the release cadence of the agent binary? how much time should we estimate to release the coupled agent binary changes?

paulyeo21: I was able to catch up with @abaptiste about his recommendation. the next step is to discuss with @abaptiste and @ar4096 about the potential solutions and come to a consensus. i will follow-up in this thread with the decision we end up with.

pkg/inject/envoy.go
paulyeo21: i understand that the networking is currently "broken" and we want to make this feature's behavior as the default, but i would still prefer to leave the default to be existing behavior and have the changes introduced in this PR to be toggle-able since this introduces a lot of change

config/helm/appmesh-controller/README.md
```diff
@@ -390,8 +390,11 @@ Parameter | Description | Default
 `sidecar.resources.requests` | Envoy container resource requests | `requests: cpu 10m memory 32Mi`
 `sidecar.resources.limits` | Envoy container resource limits | `limits: cpu "" memory ""`
 `sidecar.lifecycleHooks.preStopDelay` | Envoy container PreStop Hook Delay Value | `20s`
+`sidecar.lifecycleHooks.postStartInterval` | Envoy container PostStart Hook Interval Value | `5s`
+`sidecar.lifecycleHooks.postStartTimeout` | Envoy container PostStart Hook Timeout Value | `180s`
 `sidecar.probes.readinessProbeInitialDelay` | Envoy container Readiness Probe Initial Delay | `1s`
 `sidecar.probes.readinessProbePeriod` | Envoy container Readiness Probe Period | `10s`
+`sidecar.waitForProxyReady` | Enable pod postStart hook to delay application startup until proxy is ready to accept traffic | `false`
```
joesbigidea, line 397: This looks incorrect, the parameter in deployment.yaml is waitUntilProxyReady

Makefile
```diff
@@ -57,7 +58,7 @@ helm-lint:
 
 
 helm-deploy: check-env manifests
-	helm upgrade -i appmesh-controller config/helm/appmesh-controller --namespace appmesh-system --set image.repository=$(REPO) --set image.tag=$(VERSION) --set preview=$(PREVIEW)
+	helm upgrade -i appmesh-controller config/helm/appmesh-controller --namespace appmesh-system --set image.repository=$(REPO) --set image.tag=$(VERSION) --set preview=$(PREVIEW) --set sidecar.waitUntilProxyStarts=${WAIT_PROXY_READY}
```
joesbigidea, line 61: Shouldn't this be waitUntilProxyReady?

joesbigidea (commented):
The argument names don't seem to match up correctly

Makefile
paulyeo21: yes, thanks for catching that

config/helm/appmesh-controller/README.md
paulyeo21: ditto, like above, thanks

pkg/inject/envoy_test.go
```diff
@@ -2696,6 +2696,140 @@ func Test_envoyMutator_mutate(t *testing.T) {
...
+									Name:          "stats",
+									ContainerPort: 9901,
+									Protocol:      "TCP",
+								},
+							},
+							Lifecycle: &corev1.Lifecycle{
+								PostStart: &corev1.Handler{
+									Exec: &corev1.ExecAction{Command: []string{
+										"sh", "-c", "if [ -f /tmp/agent.sock ]; then APPNET_AGENT_POLL_ENVOY_READINESS_TIMEOUT_S=60 " +
+											"APPNET_AGENT_POLL_ENVOY_READINESS_INTERVAL_S=5 /usr/bin/agent -envoyReadiness; fi",
```
joesbigidea, line 2749: what are the cases where /tmp/agent.sock might not exist?

test/integration/sidecar-v1.22/sidecar_stack.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func (s *SidecarStack) cleanup(ctx context.Context, f *framework.Framework) {
+	if err := f.K8sClient.Delete(ctx, s.namespace); err != nil {
+		f.Logger.Error("failed to delete namespace")
+	}
+
+	if err := f.K8sClient.Delete(ctx, s.mesh); err != nil {
+		f.Logger.Error("failed to delete mesh")
+	}
```
joesbigidea, line 190: Doesn't the VN need to be deleted as well?

pkg/inject/sidecar_builder.go
```diff
@@ -190,6 +193,16 @@ func buildEnvoySidecar(vars EnvoyTemplateVariables, env map[string]string) (core
 		},
 	}
 
+	if vars.WaitUntilProxyReady {
+		envoy.Lifecycle.PostStart = &corev1.Handler{
+			Exec: &corev1.ExecAction{Command: []string{
+				"sh", "-c", fmt.Sprintf("if [ -f /tmp/agent.sock ]; then APPNET_AGENT_POLL_ENVOY_READINESS_TIMEOUT_S=%d "+
```
paulyeo21, line 199: add comment about why we need to check this /tmp/agent.sock file

paulyeo21: explore other avenues for checking envoy/agent version since agent socket file could not exist when postStart is run

paulyeo21: add an `else` echoing version of envoy is not supported with postStart

pkg/inject/envoy_test.go
paulyeo21: discussed offline -- we decided to use the envoy version to indicate that postStart hook should only run on version 1.23+

aziv2010:
What is the status of this review? It would help allot for us to use the AppMesh with our product. Thanks.

test/integration/sidecar-v1.22/sidecar_stack.go
paulyeo21: i didnt need to so i assumed they were getting deleted when deleting the mesh/namespace, but ill add it bc it seems like better practice to explicitly delete those

pkg/inject/sidecar_builder.go
```diff
@@ -190,6 +195,17 @@ func buildEnvoySidecar(vars EnvoyTemplateVariables, env map[string]string) (core
 		},
 	}
 
+	if vars.WaitUntilProxyReady {
+		envoy.Lifecycle.PostStart = &corev1.Handler{
+			Exec: &corev1.ExecAction{Command: []string{
+				"sh", "-c", fmt.Sprintf("if [[ $(/usr/bin/envoy --version) =~ ([0-9]+)\\.([0-9]+)\\.([0-9]+)-appmesh\\.([0-9]+) && "+
+					"${BASH_REMATCH[1]} -gt 2 || ${BASH_REMATCH[2]} -gt 22 || (${BASH_REMATCH[3]} -eq 2 && ${BASH_REMATCH[4]} -gt 0) ]]; "+
```
joesbigidea, line 202: Is it possible for there to be a 1.22.3 release? I see 1.21.4 came out after 1.22.2.

pkg/inject/sidecar_builder.go
```diff
@@ -190,6 +195,17 @@ func buildEnvoySidecar(vars EnvoyTemplateVariables, env map[string]string) (core
 		},
 	}
 
+	if vars.WaitUntilProxyReady {
+		envoy.Lifecycle.PostStart = &corev1.Handler{
+			Exec: &corev1.ExecAction{Command: []string{
+				"sh", "-c", fmt.Sprintf("if [[ $(/usr/bin/envoy --version) =~ ([0-9]+)\\.([0-9]+)\\.([0-9]+)-appmesh\\.([0-9]+) && "+
```
joesbigidea, line 201: This post start hook could use a comment explaining what some of the bash stuff is doing

pkg/inject/sidecar_builder.go
paulyeo21: im not sure why `1.21.4.0` is listed as after `1.22.2.0` but the most recent version before `1.22.2.1` is suppose to be `1.22.2.0` if it doesnt follow incrementing semantic versioning... then it becomes much harder to maintain imo. but regardless i think you're right we should change this to:

```
${BASH_REMATCH[3]} -gt 2 || ${BASH_REMATCH[4]} -gt 0
```

what do you think joe?

paulyeo21: talked offline-- joe let me know some examples where my condition would not work as intended to only support v1.22.2.1+. should be addressed now

paulyeo21:
> What is the status of this review? It would help allot for us to use the AppMesh with our product. Thanks.

Hey @aziv2010 the team is actively working on this. I can update this thread when it is released. Best.

joesbigidea (approved):

paulyeo21:
> What is the status of this review? It would help allot for us to use the AppMesh with our product. Thanks.

hey @aziv2010 this feature has been released on v1.6.0 now. pls try it out by enabling the feature using the following flag:

`--set sidecar.waitUntilProxyReady=true`

aziv2010:
Thanks Paul. Will do.

Avi

בתאריך יום ו׳, 19 באוג׳ 2022, 21:04, מאת Paul Yeo ***@***.***

PR #599, opened 2022-05-31
Include all /test/integration directories in script to run tests

paulyeo21:
*Issue #, if available:*

*Description of changes:*

* Include all integration tests in build process


scripts/test-with-kind.sh
```diff
@@ -71,14 +71,23 @@ function install_controller {
...
+function run_integration_tests {
+  local __vpc_id=$( vpc_id )
+
+  for __type in ${INT_TEST_DIR}/*
+  do
+    case `basename $__type` in
+      test_app) # /test_app contains test app images
+        continue
+        ;;
+    esac
```
paulyeo21, line 83: using a switch statement here so that when we have specific tests like the one included here https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/598 we can upgrade the controller with the feature sets required, like such:


```
sidecar_test)
    helm upgrade -i appmesh-controller config/helm/appmesh-controller etc etc
    ;;
```

PR #601, opened 2022-06-22
Update containerd version

bendu:
*Issue #, if available:*

*Description of changes:*
Updating containerd (dependency of helm) from 1.5.7 to 1.5.10. Updating helm (which is what's pulling in containerd) leads to a ton of other changes, including Kubernetes API changes that are not yet available until 1.23. [Branch with helm upgrade if you are curious](https://github.com/bendu/aws-app-mesh-controller-for-k8s/tree/patch)

Since EKS still supports creating K8s clusters as far back as 1.19, it did not make sense to support only 1.23+.


bendu:
Integration test is failing due to kind cluster failing on the self hosted runner.

joesbigidea (approved):

PR #603, opened 2022-06-28
Cleanup kind cluster even if integration test job fails

paulyeo21:
*Issue #, if available:*
N/A

*Description of changes:*
Make cleaning up kind cluster independent of integration test job

*Testing:*
- [x] Run commit that breaks integration test job and validate cleanup still happens

https://github.com/aws/aws-app-mesh-controller-for-k8s/runs/7101608920?check_suite_focus=true

- [x] Confirm kind clusters are clearing properly
Screen Shot 2022-06-28 at 15 57 00

### DIFF
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -15,6 +15,9 @@ jobs:
     name: integration-test
     runs-on: [self-hosted, aws-app-mesh-controller-for-k8s, X64 ]
     steps:
+      - name: intentionally break
+        run: |
+          lkasdf
       - name: clean work dir from previous runs
         run: |
           rm -rf *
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-test.yaml
```diff
@@ -30,5 +33,10 @@ jobs:
         uses: actions/checkout@v2
       - name: setup kind and run integration tests
         run: make integration-test
-      - name: cleanup all the kind clusters
+  cleanup:
+    runs-on: [self-hosted, aws-app-mesh-controller-for-k8s, X64]
+    if: ${{ always() }}
+    needs: [build]
+    steps:
+      - name: delete kind clusters
```
joesbigidea, line 41: Is it possible to have multiple test runs going at once? If so I would imagine this would break the other runs.

paulyeo21: @jeremymill @bendu do either of you have a definitive answer to joe's q?

bendu: Only 1 job runs on the host at a time. They queue up if the host is busy

bendu (approved):

bendu:
sheep it

PR #612, opened 2022-07-20
Push controller image to new opt-in region CGK

paulyeo21:
*Issue #, if available:*

N/A

*Description of changes:*

Support opt-in region CGK in controller release

- [x] Configure secrets with CGK role and acccount
> https://github.com/aws/aws-app-mesh-controller-for-k8s/settings/secrets/actions
- [x] Ensure ECR repo exists
>  Screen Shot 2022-07-20 at 07 12 27

### DIFF
--- a/.github/workflows/prod-release.yaml
+++ b/.github/workflows/prod-release.yaml
@@ -101,7 +101,7 @@ jobs:
           role-session-name: ControllerProdRelease
       - name: Push docker images (CGK)
         run: |
-          aws ecr get-login-password --region eu-south-1 | docker login --username AWS --password-stdin ${{ secrets.PROD_CGK_AWS_ACCOUNT }}.dkr.ecr.ap-southeast-3.amazonaws.com
+          aws ecr get-login-password --region ap-southeast-3 | docker login --username AWS --password-stdin ${{ secrets.PROD_CGK_AWS_ACCOUNT }}.dkr.ecr.ap-southeast-3.amazonaws.com
           docker tag ${{ secrets.BETA_AWS_ACCOUNT }}.dkr.ecr.us-west-2.amazonaws.com/amazon/appmesh-controller:${{ github.event.inputs.tag }} ${{ secrets.PROD_CGK_AWS_ACCOUNT }}.dkr.ecr.ap-southeast-3.amazonaws.com/amazon/appmesh-controller:${{ github.event.inputs.tag }}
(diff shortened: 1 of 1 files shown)


.github/workflows/prod-release.yaml
```diff
@@ -93,6 +93,19 @@ jobs:
           docker tag ${{ secrets.BETA_AWS_ACCOUNT }}.dkr.ecr.us-west-2.amazonaws.com/amazon/appmesh-controller:${{ github.event.inputs.tag }}-linux_arm64 ${{ secrets.PROD_MXP_AWS_ACCOUNT }}.dkr.ecr.eu-south-1.amazonaws.com/amazon/appmesh-controller:${{ github.event.inputs.tag }}-linux_arm64
           docker push ${{ secrets.PROD_MXP_AWS_ACCOUNT }}.dkr.ecr.eu-south-1.amazonaws.com/amazon/appmesh-controller:${{ github.event.inputs.tag }}
           docker push ${{ secrets.PROD_MXP_AWS_ACCOUNT }}.dkr.ecr.eu-south-1.amazonaws.com/amazon/appmesh-controller:${{ github.event.inputs.tag }}-linux_arm64
+      - name: Configure AWS Credentials (prod-CGK)
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-region: us-west-2
```
paulyeo21, line 99: this is slightly confusing that the region here is `us-west-2` but this is how we're doing it for all the other regions. @jeremymill i assume we do this because this is part of the initial tag we generate when we build the image?

.github/workflows/prod-release.yaml
```diff
@@ -93,6 +93,19 @@ jobs:
...
+          role-to-assume: ${{ secrets.PROD_CGK_AWS_ROLE }}
+          role-session-name: ControllerProdRelease
+      - name: Push docker images (CGK)
+        run: |
+          aws ecr get-login-password --region eu-south-1 | docker login --username AWS --password-stdin ${{ secrets.PROD_CGK_AWS_ACCOUNT }}.dkr.ecr.ap-southeast-3.amazonaws.com
```
joesbigidea, line 104: --region looks incorrect here

.github/workflows/prod-release.yaml
bendu: The story behind this is that we had problems with the STS endpoint for af-south-1, but creds are not region specific, so we just get creds from us-west-2 regardless of which region we actually need to use.

bendu: Of course, china regions and any other aws partition would need to be in their respective region.

.github/workflows/prod-release.yaml
paulyeo21: ah yeah, thanks

.github/workflows/prod-release.yaml
paulyeo21: gotcha thx for the context

jeremymill (approved):

joesbigidea (approved):

PR #613, opened 2022-07-21
Add Backend Groups

jeremymill:
*Issue #, if available:*

*Description of changes:*
Adds a new BackendGroup resource. A Backend Group specifies a list of Virtual Service backends, which can be re-used across multiple Virtual Nodes.

A VirtualNode can reference a Backend Group by its Namespace and Name, similar to a Virtual Service backend. Specifying the name as "*" will add all Virtual Services in the specified namespace to the Virtual Node. Any future Virtual Services created in that namespace will be added to the Virtual Node automatically.

Note: BackendGroup is intended to be a v1alpha1 resource, but it requires v1beta2 spec changes so that Virtual Nodes can reference the groups. An alternative strategy may be to offer this feature in a beta release of the controller.


test/framework/resource/virtualnode/manager.go
```diff
@@ -113,7 +115,9 @@ func (m *defaultManager) CheckVirtualNodeInAWS(ctx context.Context, ms *appmesh.
 	}
 	opts := equality.CompareOptionForVirtualNodeSpec()
 	if !cmp.Equal(desiredSDKVNSpec, resp.VirtualNode.Spec, opts) {
-		return errors.New(cmp.Diff(desiredSDKVNSpec, resp.VirtualNode.Spec, opts))
+		return fmt.Errorf("Observed:\n%v\n\nExpected:\n%v\nvsRefs:\n%v\nExpected VN:\n%v",
+			resp.VirtualNode.Spec.Backends, desiredSDKVNSpec.Backends, vsRefs, vn.Spec.Backends)
+		//return errors.New(cmp.Diff(desiredSDKVNSpec, resp.VirtualNode.Spec, opts))
```
joesbigidea, line 120: seems safe to delete the commented out code

pkg/virtualnode/enqueue_requests_for_backendgroup_events.go
```diff
@@ -0,0 +1,68 @@
...
+	bgOld := e.ObjectOld.(*appmesh.BackendGroup)
+	bgNew := e.ObjectNew.(*appmesh.BackendGroup)
+	if !reflect.DeepEqual(bgOld.Spec.VirtualServices, bgNew.Spec.VirtualServices) {
+		h.enqueueVirtualNodesForMesh(context.Background(), queue, bgNew.Spec.MeshRef)
+	}
+}
+
+// Delete is called in response to a delete event
+func (h *enqueueRequestsForBackendGroupEvents) Delete(e event.DeleteEvent, queue workqueue.RateLimitingInterface) {
+	// no-op
```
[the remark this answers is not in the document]
jeremymill: I agree in terms of the k8s model, though this isn't unique to backend groups as currently virtual nodes don't watch for virtual service deletion.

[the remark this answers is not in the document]
joesbigidea: I think the impact of this might be pretty different in that for backend groups you end up with VN instances that still have the previously associated services as backends. In the case of a VS being deleted I don't think the Envoy will get that service as a backend regardless of the fact that it's still configured on the VN. I'm open to whether that's really a concerning difference though. Thoughts?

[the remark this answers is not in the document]
jeremymill: Currently if a VN has a VS reference that can't be found, the reconcile attempt will error out and get re-queued. I followed that same pattern with backend groups. To get rid of backends from deleted backend groups, we'd need to be able to reconcile the VN while ignoring the deleted backend groups. My concern here is this leaves room to miss backend groups if the VN can't find them for some reason.

pkg/virtualnode/enqueue_requests_for_backendgroup_events.go
```diff
@@ -0,0 +1,75 @@
...
+		h.log.Error(err, "failed to enqueue virtualNodes for backend group events",
+			"mesh", meshRef.Name)
+		return
+	}
+	for _, vn := range vnList.Items {
+		if vn.Spec.MeshRef == nil || *meshRef != *vn.Spec.MeshRef {
+			continue
+		}
+		for _, bg := range vn.Spec.BackendGroups {
+			if bg.Name == backendGroup.Name && bg.Namespace != nil && *bg.Namespace == backendGroup.Namespace {
```
joesbigidea, line 69: Is this supposed to be (bg.Namespace != nil || *bg.Namespace == backendGroup.Namespace)?

joesbigidea, line 69: If bg.Namespace == nil then shouldn't it check vn.Namespace == backendGroup.Namespace? That's how it's matching when processing the VN right?

config/samples/appmesh_v1beta2_backendgroup.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: appmesh.k8s.aws/v1beta2
+kind: BackendGroup
+metadata:
+  name: backendgroup-sample
+spec:
+  # TODO(user): Add fields here
```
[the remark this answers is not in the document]
jeremymill: None of the samples seem to be used (here's config/samples/appmesh_v1beta2_virtualnode.yaml)
```
kind: VirtualNode
metadata:
  name: virtualnode-sample
spec:
  # Add fields here
  foo: bar
```

Dockerfile
```diff
@@ -8,6 +8,7 @@ WORKDIR /workspace
 COPY . ./
 # cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
+ENV GOPROXY direct
```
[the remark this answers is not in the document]
jeremymill: nope, should be removed now

pkg/virtualnode/enqueue_requests_for_backendgroup_events.go
```diff
@@ -0,0 +1,68 @@
...
+	// no-op
+}
+
+// Generic is called in response to an event of an unknown type or a synthetic event triggered as a cron or
+// external trigger request
+func (h *enqueueRequestsForBackendGroupEvents) Generic(e event.GenericEvent, queue workqueue.RateLimitingInterface) {
+	// no-op
+}
+
+func (h *enqueueRequestsForBackendGroupEvents) enqueueVirtualNodesForMesh(ctx context.Context, queue workqueue.RateLimitingInterface, meshRef *appmesh.MeshReference) {
```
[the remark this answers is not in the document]
jeremymill: updated

pkg/virtualnode/enqueue_requests_for_virtualservice_events.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// Generic is called in response to an event of an unknown type or a synthetic event triggered as a cron or
+// external trigger request
+func (h *enqueueRequestsForVirtualServiceEvents) Generic(e event.GenericEvent, queue workqueue.RateLimitingInterface) {
+	// no-op
+}
+
+func (h *enqueueRequestsForVirtualServiceEvents) enqueueVirtualNodesForMesh(ctx context.Context, queue workqueue.RateLimitingInterface, meshRef *appmesh.MeshReference) {
+	vnList := &appmesh.VirtualNodeList{}
```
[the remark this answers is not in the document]
jeremymill: updated

pkg/virtualnode/resource_manager.go
```diff
@@ -133,8 +134,35 @@ func (m *defaultResourceManager) validateMeshDependencies(ctx context.Context, m
...
 func (m *defaultResourceManager) findVirtualServiceDependencies(ctx context.Context, vn *appmesh.VirtualNode) (map[types.NamespacedName]*appmesh.VirtualService, error) {
-	vsByKey := make(map[types.NamespacedName]*appmesh.VirtualService, len(vn.Spec.Backends))
+	vsByKey := make(map[types.NamespacedName]*appmesh.VirtualService)
 	vsRefs := ExtractVirtualServiceReferences(vn)
+	for _, backendGroupRef := range vn.Spec.BackendGroups {
+		// Wildcard special case
+		bgKey := references.ObjectKeyForBackendGroupReference(vn, backendGroupRef)
+		if bgKey.Name == "*" {
+			m.log.Info("Wildcard spotted!")
+			// TODO get all virtual services in bgKey.Namespace
```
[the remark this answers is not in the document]
jeremymill: fixed

pkg/virtualnode/resource_manager.go
```diff
@@ -133,8 +134,35 @@ func (m *defaultResourceManager) validateMeshDependencies(ctx context.Context, m
...
+				return nil, fmt.Errorf("could not list virtualservices for namespace: %s", bgKey.Namespace)
+			}
+			for _, vs := range vsList.Items {
+				vsRef := appmesh.VirtualServiceReference{
+					Namespace: aws.String(vs.Namespace),
+					Name:      vs.Name,
+				}
+				vsRefs = append(vsRefs, vsRef)
+			}
+			continue
```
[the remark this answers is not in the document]
jeremymill: updated

pkg/virtualnode/resource_manager.go
```diff
@@ -311,7 +343,19 @@ func BuildSDKVirtualNodeSpec(vn *appmesh.VirtualNode, vsByKey map[types.Namespac
 		return sdkVSRefConvertFunc(a.(*appmesh.VirtualServiceReference), b.(*string), scope)
 	})
 	sdkVNSpec := &appmeshsdk.VirtualNodeSpec{}
-	if err := converter.Convert(&vn.Spec, sdkVNSpec, nil); err != nil {
+	newSpec := vn.Spec.DeepCopy()
```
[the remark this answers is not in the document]
jeremymill: updated

test/framework/resource/virtualnode/manager.go
jeremymill: fixed, made a new error message just for debugging but not looking to push it as a change

pkg/virtualnode/enqueue_requests_for_backendgroup_events.go
jeremymill: should be fixed

joesbigidea (approved):
Looks great!

pkg/virtualnode/resource_manager.go
```diff
@@ -196,10 +222,14 @@ func (m *defaultResourceManager) createSDKVirtualNode(ctx context.Context, ms *a
 
 func (m *defaultResourceManager) updateSDKVirtualNode(ctx context.Context, sdkVN *appmeshsdk.VirtualNodeData, ms *appmesh.Mesh, vn *appmesh.VirtualNode, vsByKey map[types.NamespacedName]*appmesh.VirtualService) (*appmeshsdk.VirtualNodeData, error) {
 	actualSDKVNSpec := sdkVN.Spec
+	for i, j := range vsByKey {
+		m.log.Info(vn.Name + "> " + i.String() + ", " + j.Name)
```
paulyeo21, line 226: why do we log this and in this format? `virtual node name > [index], virtual service key`

pkg/virtualnode/resource_manager.go
```diff
@@ -311,7 +341,19 @@ func BuildSDKVirtualNodeSpec(vn *appmesh.VirtualNode, vsByKey map[types.Namespac
 		return sdkVSRefConvertFunc(a.(*appmesh.VirtualServiceReference), b.(*string), scope)
 	})
 	sdkVNSpec := &appmeshsdk.VirtualNodeSpec{}
-	if err := converter.Convert(&vn.Spec, sdkVNSpec, nil); err != nil {
+	tempSpec := vn.Spec.DeepCopy()
```
paulyeo21, line 344: dont we get this for free? since existing behavior is that we build the virtual service sdk reference for backends i.e. https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/613/files#diff-93492cb609ae37c6f1dc762777019e79a9eb10d2d20cd6e0c716c776829eafe7R339

pkg/virtualnode/resource_manager.go
```diff
@@ -133,8 +134,33 @@ func (m *defaultResourceManager) validateMeshDependencies(ctx context.Context, m
...
+			bg, err := m.referencesResolver.ResolveBackendGroupReference(ctx, vn, backendGroupRef)
+			if err != nil {
+				return nil, errors.Wrapf(err, "failed to resolve backendGroupRef")
+			}
+			vsRefs = append(vsRefs, bg.Spec.VirtualServices...)
+		}
+	}
 	for _, vsRef := range vsRefs {
 		vsKey := references.ObjectKeyForVirtualServiceReference(vn, vsRef)
 		if _, ok := vsByKey[vsKey]; ok {
```
paulyeo21, line 166: this is nice that the existing logic does a map here since we dont have to dedupe virtual services if customers configure overlapping backendgroups

pkg/virtualnode/resource_manager.go
jeremymill: debugging info I left in there, will remove

pkg/virtualnode/resource_manager.go
jeremymill: It does do the conversion, but it does it based on the Backends field of the VirtualNodeSpec, not vsByKey. So we don't get the new backends for free.

paulyeo21: maybe im missing something but dont we throw all the virtual services from the backend group into the map `vsByKey` and the existing conversion function https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/references/conversions.go#L54 uses that map?

jeremymill: Not quite, that conversion function defines conversions for all the vs in vsByKey, but only the backends from the virtual node spec are actually passed as inputs to that function: https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/conversions/virtualnode_types_conversion.go#L647 I've tested without the tempSpec code and the backends from backend groups do not make it into the SDK spec.

paulyeo21: i see i see reading deeper into the conversion logic i realized the virtual service conversion actually happens here https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/conversions/virtualnode_types_conversion.go#L142 its just registered earlier in this function. i guess the problem i have with adding this logic here is that we have specific functions for converting CRD resources to SDK resources. And in your case i realize you dont have a strict one-to-one SDK resource, but you are converting to SDK backend resources. since the CRD backgroup spec has virtual service references, can we not loop through the virtual services say here https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/conversions/virtualnode_types_conversion.go#L654 and add it to `sdkObj.Backends` ? it also seems like even though we dedupe the virtual service references in the convert func since we use a map, we dont actually do that in the actual conversion since we're appending all backends from backend groups regardless of whether those backends were already provided. i know this is the last thing holding up this review so lmk if you want to chat offline

jeremymill: The tempSpec's backends are set to (not appended to) the already de-duped list containing both explicitly defined backends and ones from groups (vsByKey), so there shouldn't be any additional de-duping issue in the conversion step.

I do understand that having the conversion functions effectively ignore backend groups isn't ideal, but moving backendgroup processing into the conversion functions would require passing in a referenceResolver, or at least a k8sClient, in order to get the BackendGroup objects (whose specs contain the VS lists we want to loop through). I don't love this because there's a strong convention that the inputs to those conversion functions are crdObj, sdkObj, scope. Let's go ahead and sync up offline about this.

paulyeo21: talked offline-- jeremy walked me through what he mentioned above. LGTM

paulyeo21 (approved):
awesome work!

scripts/test-with-kind.sh
```diff
@@ -63,7 +63,7 @@ function install_controller {
        local __controller_name="appmesh-controller"
        local __ns="appmesh-system"
        kubectl create ns $__ns
-       APPMESH_PREVIEW=y AWS_ACCOUNT=$AWS_ACCOUNT_ID AWS_REGION=$AWS_REGION make helm-deploy
+       APPMESH_PREVIEW=y AWS_ACCOUNT=$AWS_ACCOUNT_ID AWS_REGION=$AWS_REGION ENABLE_BACKEND_GROUPS=true make helm-deploy
```
paulyeo21, line 66: this will enable the backend groups for all tests (which doesnt sound bad tbh) but if you wanted to gate this for only your integration test suite you can do a separate helm-deploy as a case on L80 like i do in my PR here https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/598/files#diff-af7f07af62d09eec6acc7e70d0f073dcd4b6326b52e02853b2adc6a8984fe73eR90

paulyeo21 (approved):

joesbigidea (approved):

PR #622, opened 2022-08-09
Account id handling

jgobrien09:
Added a function to check whether an AccountID is formatted correctly, in the event the AccountID was converted to scientific notation, convert the ID back to a 12 digit string value.

Added unit tests to verify behavior.


joesbigidea (approved):
looks good

pkg/aws/cloud_config.go
```diff
@@ -25,3 +28,21 @@ func (cfg *CloudConfig) BindFlags(fs *pflag.FlagSet) {
...
+	if properIDMatched || cfg.AccountID == "" {
+		return
+	}
+	log.Error(nil, "The following AWS Account ID is not formatted correctly: "+cfg.AccountID)
+
+	scientificMatched, _ := regexp.MatchString("^(\\d[.])(\\d{11})(e\\+11)$", cfg.AccountID)
+	if scientificMatched {
+		cfg.AccountID = cfg.AccountID[0:13]
+		cfg.AccountID = strings.Replace(cfg.AccountID, ".", "", 1)
+		log.Error(nil, "Using the converted AWS Account ID: "+cfg.AccountID)
```
paulyeo21, line 46: nit: `log.Error(nil, fmt.Sprintf("Using the converted AWS Account ID: %s", cfg.AccountID))`

pkg/aws/cloud_config_test.go
```diff
@@ -0,0 +1,98 @@
...
+	cloudConfig := CloudConfig{
+		Region:         "",
+		AccountID:      "",
+		ThrottleConfig: nil,
+	}
+	setupLog := ctrl.Log.WithName("setup")
+
+	type args struct {
+		accountId string
+	}
```
paulyeo21, line 19: nit: i would get rid of this since imo it doesnt add much

```
tests: []struct {
    name string
    wantAccountID string
    accountID string
}{
...
}
...

for _, tt := range tests {
    t.Run(tt.name, func(t *testing.T) {
        cloudConfig.AccountID = tt.accountId
...
        })
    }
```

paulyeo21 (approved):
LGTM, thanks for writing the table tests!

pkg/aws/cloud_config_test.go
jgobrien09: isn't that what runs all the unit tests? Just assuming based on other test classes I found

paulyeo21: im not sure what you mean by it runs all the unit tests? or do you mean thats how we're doing in the other tests? yeah this is how the other tests do it, but imo its not necessary. all it does is create a nested object inside the test object

```
tests := []struct {
    name string
    args args
    wantAccountID string
}
```

test here is an unamed struct but you could name it like such: 

```
type test struct {
    name string
    args args
    wantAccountID string
}
```

your `args` here is your "arguments" to your tests, but imo its clear that the one argument you need is the account id so it doesnt add that much value for how much boilerplate it requires (like doing `args: args{acccountId:...}` everywhere

that way we can have something like

```
{
    name: "normal case",
    wantAccountID: "123456789012",
    acccountId: "123456789012",
}
{
...
}
```

pkg/aws/cloud_config.go
jgobrien09: fixed thanks Paul

pkg/aws/cloud_config_test.go
jgobrien09: fixed thanks for clarifying I was thinking you were saying to pull out the whole testing looping structure, but I got you. args removed

PR #634, opened 2022-09-07
Helm release v1.7.0

paulyeo21:
*Issue #, if available:*

*Description of changes:*

Helm release v1.7.0

Includes:

Multiple listener support

### DIFF
--- a/config/helm/appmesh-controller/ci/values.yaml
+++ b/config/helm/appmesh-controller/ci/values.yaml
@@ -4,6 +4,6 @@
 accountId: 123456789
 region: us-west-2
 image:
-  repository: jeremyymill/appmesh-controller
+  repository: paulyeo21/appmesh-controller
   tag: v1.7.0
   pullPolicy: IfNotPresent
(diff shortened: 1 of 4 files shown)


config/helm/appmesh-controller/ci/values.yaml
```diff
@@ -5,5 +5,5 @@ accountId: 123456789
 region: us-west-2
 image:
   repository: jeremyymill/appmesh-controller
```
jeremymill, line 7: so this is a weird thing where the CI on eks-charts actually pulls from dockerhub instead of ECR; current workaround is to upload the image to dockerhub and put the repo name and tag here

paulyeo21: oh i see what you mean. ok so i should push to my dockerhub with the proper tag and change the repository value here?

paulyeo21: Screen Shot 2022-09-07 at 12 38 23

jeremymill (approved):

PR #640, opened 2022-09-23
[DRAFT] Move integration tests to EKS

jeremymill:
*Issue #, if available:*

*Description of changes:*
This PR aims to move integration tests off of self-hosted runners by running them in EKS.

Currently, most tests pass, but the VirtualNode CloudMap tests are panicking due to the private DNS namespaces not being created for some reason. The API calls appear in CloudTrail and there don't appear to be errors, so I'm not sure what the issue is.

Current debug stuff:
- two fmt.Println debug statements
- integration test loop is ignored and only runs virtualnode tests (trying to isolate the cloudmap issue)


bendu:
We got github actions hosted kind tests working in #667

PR #644, opened 2022-10-17
Adding the ability to choose fips and dualstack endpoints

sshver:
*Issue #, if available:*

*Description of changes:*
- Adding the ability to choose FIPS and DualStack endpoints for App Mesh.


pkg/inject/config.go
```diff
@@ -205,6 +209,8 @@ func (cfg *Config) BindFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&cfg.DualStackEndpoint, flagDualStackEndpoint, false, "Use DualStack Endpoint")
 	fs.BoolVar(&cfg.DualStackEndpoint, flagEnvoyAdminAccessEnableIpv6, false, "Enable Admin access when using IPv6")
 	fs.StringVar(&cfg.ClusterName, flagClusterName, "", "ClusterName in context")
+	fs.IntVar(&cfg.UseFipsEndpoint, flagUseFipsEndpoint, 0, "To use FIPS endpoint")
+	fs.IntVar(&cfg.UseDualStack, flagUseDualStack, 0, "To use Dual Stack Endpoint")
```
joesbigidea, line 213: There's already a DualStackEndpoint with an identical description above (then another flag bound to the same flag for enable IPv6 admin endpoint). I see that it's used in the Envoy sidecar configuration. How do these decisions relate to each other? Does it make sense to make a separate flag (might be necessary for backward compat)? If these should be separate I think they need their descriptions updated a bit to make them easier to understand and maybe update your new flag name to differentiate it.

pkg/inject/config.go
```diff
@@ -112,7 +114,9 @@ type Config struct {
 	StatsDSocketPath     string
 	XRayImage            string
 
-	ClusterName string
+	ClusterName     string
+	UseFipsEndpoint int
+	UseDualStack    int
```
joesbigidea, line 119: These seem like they should be bool instead of int. That's also consistent with the AWS_USE_FIPS_ENDPOINT env var that pops up other places.

bendu:
Can we remove the DS Store files and add them to .gitignore?

pkg/aws/cloud.go
```diff
@@ -66,7 +76,7 @@ func NewCloud(cfg CloudConfig, metricsRegisterer prometheus.Registerer) (Cloud,
 		cfg:      cfg,
 		appMesh:  services.NewAppMesh(sess),
 		cloudMap: services.NewCloudMap(sess),
-		eks:      services.NewEKS(sess),
+		eks:      services.NewEKS(sessEKS),
```
bendu, line 79: I know EKS does not have dual stack endpoints, but it does have FIPS. This does not seem like it should ignore the FIPS endpoints

bendu: Can we create a separate config for EKS that can set FIPS but not dual stack?

BennettJames:
This is minor, but can the `.DS_Store` files be removed from the commit? And/or add `.DS_Store` to `.gitignore`.

bendu:
Succeeded by Dileep's changes

Issue #646, opened 2022-10-21
Add support for configuring APPMESH_ENABLE_IPV6 env var in proxyinit container


andrewmarklloyd:
**Is your feature request related to a problem?**
Our EKS clusters have IPV6 disabled so the proxy init container on our pods fail to run because the `APPMESH_ENABLE_IPV6` env var is [hard coded](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/inject/init_proxy.go#L50-L51) to `1`.

Here is the error message we see in the init container:
```
ip6tables v1.8.4 (legacy): can't initialize ip6tables table `nat': Address family not supported by protocol
Perhaps ip6tables or your kernel needs to be upgraded.
```

**Describe the solution you'd like**
Similar to how [annotations](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/inject/constants.go) on the pod are exposed to configure options on the proxyinit container, I suggest adding an annotation to set `APPMESH_ENABLE_IPV6=0`.

**Describe alternatives you've considered**
We've investigated building our own image setting APPMESH_ENABLE_IPV6=0 in the entrypoint. This works but requires us to manage the image lifecycle instead of relying on updates to the source image.

For example:
```
FROM 840364872350.dkr.ecr.us-west-2.amazonaws.com/aws-appmesh-proxy-route-manager:v6-prod
ENTRYPOINT [ "/bin/sh", "-c", "APPMESH_ENABLE_IPV6=0", "/opt/amazon/bin/sidecar-proxy-route-manager.sh" ]
```

andrewmarklloyd:
I see that #647 was merged which is great! Thank you for allowing me to contribute to the project, it was really cool.
Should I close this issue or is that up to the maintainers? Also what is the deployment process for the controller? Will that change be incorporated in the `v6-prod` tag or will another one be created?

bendu:
Hey @andrewmarklloyd I have closed the issue. Thank you for contributing!

[...]

PR #647, opened 2022-10-24
Feature/allow ipv6 disable annotation

andrewmarklloyd:
*Issue #, if available:*
Implements the feature described in #646

*Description of changes:*
Similar to how [annotations](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/pkg/inject/constants.go) on the pod are exposed to configure options on the proxyinit container, this feature adds an annotation to set `APPMESH_ENABLE_IPV6=0` in the proxyinit container.

Adds tests to ensure the default `APPMESH_ENABLE_IPV6=1` is set when the annotation is unset as well as tests `APPMESH_ENABLE_IPV6=0` when the annotation is set to `appmesh.k8s.aws/ipv6: disabled`.

Additionally I ran `make` locally which made changes to `apis/appmesh/v1beta2/zz_generated.deepcopy.go`.


bendu:
Hi Andrew. Thank you for your PR. We will review it as soon as possible.

pkg/inject/proxy.go
```diff
@@ -56,18 +57,22 @@ type proxyConfig struct {
 	proxyIngressPort int64
 	// UID used by proxy
 	proxyUID int64
+	// todo
+	enableIPV6 *bool
```
bendu, line 61: Does this need to be a *bool? Why not just a bool?

andrewmarklloyd: I found when using a bool instead of *bool, and `enableIPV6` was not set explicitly in `proxyConfig`, that the default value was `false`. This resulted in unexpected value of `APPMESH_ENABLE_IPV6: 0` in the proxyinit env vars.

This allows checking for `m.proxyConfig.enableIPV6 != nil && !*m.proxyConfig.enableIPV6` ensuring that the default value of `APPMESH_ENABLE_IPV6` remains `1`.

bendu: Ah understood. Can you add a comment explaining this? I want to make sure the next person who looks at the code understands this is to set the default to 1 when the variable is unset.

bendu (approved):

BennettJames (approved):

PR #655, opened 2022-11-16
Allow for configuration of GORPOXY in docker build

BennettJames:
*Issue #, if available:*

*Description of changes:*

Let's the user customize the docker build GOPROXY variable when building the controller container image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ generate: controller-gen
 # Requires buildx to be installed: https://docs.docker.com/buildx/working-with-buildx/
 # By default the TARGETPLATFORM is set to linux/amd64, change the value if you are building for linux/arm64
 docker-build: check-env test
-	docker buildx build --platform linux/amd64 -t $(IMAGE) . --load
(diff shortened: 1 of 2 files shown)


rakeb (approved):

Makefile
```diff
@@ -88,7 +88,7 @@ generate: controller-gen
 # Requires buildx to be installed: https://docs.docker.com/buildx/working-with-buildx/
 # By default the TARGETPLATFORM is set to linux/amd64, change the value if you are building for linux/arm64
 docker-build: check-env test
-	docker buildx build --platform linux/amd64 -t $(IMAGE) . --load
+	docker buildx build --platform linux/amd64 --build-arg GOPROXY=$(GOPROXY) -t $(IMAGE) . --load
```
rakeb, line 91: Are we only considering `linux/amd64`?

BennettJames: The github workflow performs its own builds, including an arm64 image. Perhaps we could be a bit more clever about sharing the build setup and configuration for each, but as this is an existing setup that's not being changed by this patch I'm ok leaving that for now.

PR #656, opened 2022-11-16
Adding Envoy variable in order to use FIPS endpoint

dileepng:
*Issue #, if available:*

*Description of changes:*
Added Envoy variable which enables the feature to use FIPS Endpoint 

### DIFF
--- a/pkg/inject/envoy.go
+++ b/pkg/inject/envoy.go
@@ -52,6 +52,7 @@ type envoyMutatorConfig struct {
 	k8sVersion                 string
 	useDualStackEndpoint       bool
 	enableAdminAccessIPv6      bool
+	useFipsEndpoint            bool
 }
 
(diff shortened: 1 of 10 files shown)


bendu (approved):
ship it

joesbigidea (approved):
Looks great!

pkg/inject/envoy.go
```diff
@@ -274,3 +277,11 @@ func (m *envoyMutator) getUseDualStackEndpoint(useDualStackEndpoint bool) string
 		return "0"
 	}
 }
+
+func (m *envoyMutator) getUseFipsEndpoint(useFipsEndpoint bool) string {
+	if useFipsEndpoint {
+		return "1"
```
EdwardXF, line 283: why we pass String instead of bool here?

dileepng: Actually Controller takes boolean and it converts into String because Envoy accepts FIPS endpoint as ENV variable which is String. so we are converting and passing it to envoy. If I am not wrong.

EdwardXF: Makes sense, thanks

PR #657, opened 2022-11-17
Bump Envoy image version to v1.24.0.0-prod

ytsssun:
Issue #, if available:

Description of changes:
Bump AppMesh Envoy image version from v1.22.2.1-prod to v1.24.0.0-prod .
Ref: https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/621
 Envoy v1.24.0 release notes: https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.24/v1.24.0

Note: please monitor this https://github.com/aws/aws-app-mesh-roadmap/issues/442 and make sure the v1.24.0.0 Envoy image is available in public ECR before merging this PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ IMAGE ?= $(REPO):$(VERSION)
 PREVIEW=false
 ENABLE_BACKEND_GROUPS?=false
 WAIT_PROXY_READY=false
-SIDECAR_IMAGE_TAG=v1.22.2.1-prod
+SIDECAR_IMAGE_TAG=v1.24.0.0-prod
 
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true,crdVersions=v1"
--- a/config/helm/appmesh-controller/test.yaml
+++ b/config/helm/appmesh-controller/test.yaml
@@ -16,7 +16,7 @@ image:
 sidecar:
   image:
     repository: 840364872350.dkr.ecr.us-west-2.amazonaws.com/aws-appmesh-envoy
-    tag: v1.22.2.1-prod
+    tag: v1.24.0.0-prod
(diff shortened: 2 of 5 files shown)


suniltheta (approved):

BennettJames:
I believe the makefile should be updated as well -

- https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/Makefile#L12

ytsssun:
> I believe the makefile should be updated as well -
> 
>     * https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/Makefile#L12

Checked with @paulyeo21 offline, looks like that version indicates the beginning of the agent feature [waitUntilProxyReady](https://github.com/aws/aws-app-mesh-controller-for-k8s/commit/4a119be78dc1f3dbca88447f5169fc2dead894b9), hence should not be updated.

suniltheta:
Better to rename that variable & add comments to avoid confusion.

BennettJames:
Hrm, yeah I think we might need to think through this policy a bit more then. The integration tests as-is will call `make integration-test`, which in turn will call `make helm-deploy` which uses that variable. Having the integration tests not use the default-configured envoy does not seem appropriate to me.

BennettJames:
Yeah, having thought about this a bit more, I'm pretty sure we should increment that variable. There are still ways to test older versions by setting that variable (i.e. `make integration-test SIDECAR_IMAGE_TAG=v1.22.2.1-prod`), but as-is it acts as a default for controller deploys. The default should be consistently set to the latest. The variable in the Makefile can be downgraded if desired by a user.

I think there's a separate value check w/in the sidecar builder that performs a check against the version. That should stay the same, but I believe the makefile needs updating.

BennettJames (approved):

ytsssun:
Synced with @rakeb, I think I was wrong about the Makefile. The Makefile is only related to the Integ test. We need to include the version update. Updated the PR.

PR #658, opened 2022-11-22
Adding ability to use FIPS Endpoint for AWS Services and DUALSTACK endpoint for AWS Service(APPMESH)

dileepng:
*Issue #, if available:*

*Description of changes:*
Adding functionality to choose FIPS and DUALSTACK Endpoints to use AWS Api Services for AppMesh,EKS,CloudMap.

### DIFF
--- a/config/helm/appmesh-controller/templates/deployment.yaml
+++ b/config/helm/appmesh-controller/templates/deployment.yaml
@@ -74,6 +74,9 @@ spec:
         - --envoy-admin-access-enable-ipv6={{ .Values.sidecar.envoyAdminAccessEnableIPv6 }}
         - --dual-stack-endpoint={{ .Values.sidecar.useDualStackEndpoint }}
         - --fips-endpoint={{ .Values.sidecar.useFipsEndpoint }}
+        - --envoy-aws-access-key-id={{ .Values.sidecar.envoyAwsAccessKeyId }}
+        - --envoy-aws-secret-access-key={{ .Values.sidecar.envoyAwsSecretAccessKey }}
+        - --envoy-aws-session-token={{ .Values.sidecar.envoyAwsSessionToken }}
         - --preview={{ .Values.preview }}
         - --enable-sds={{ .Values.sds.enabled }}
         - --sds-uds-path={{ .Values.sds.udsPath }}
--- a/config/helm/appmesh-controller/templates/deployment.yaml
+++ b/config/helm/appmesh-controller/templates/deployment.yaml
@@ -79,6 +79,8 @@ spec:
         - --sds-uds-path={{ .Values.sds.udsPath }}
         - --enable-backend-groups={{ .Values.enableBackendGroups }}
         - --cluster-name={{ .Values.clusterName}}
+        - --use-aws-dual-stack-endpoint={{ .Values.useAwsDualStackEndpoint}}
+        - --use-aws-fips-endpoint={{ .Values.useAwsFIPSEndpoint}}
         {{- if .Values.cloudMapCustomHealthCheck.enabled }}
         - --enable-custom-health-check=true
         {{- end }}
--- a/config/helm/appmesh-controller/templates/deployment.yaml
+++ b/config/helm/appmesh-controller/templates/deployment.yaml
@@ -73,6 +73,7 @@ spec:
         - --envoy-admin-access-log-file={{ .Values.sidecar.envoyAdminAccessLogFile }}
         - --envoy-admin-access-enable-ipv6={{ .Values.sidecar.envoyAdminAccessEnableIPv6 }}
         - --dual-stack-endpoint={{ .Values.sidecar.useDualStackEndpoint }}
+        - --fips-endpoint={{ .Values.sidecar.useFipsEndpoint }}
         - --preview={{ .Values.preview }}
         - --enable-sds={{ .Values.sds.enabled }}
         - --sds-uds-path={{ .Values.sds.udsPath }}
--- a/config/helm/appmesh-controller/test.yaml
+++ b/config/helm/appmesh-controller/test.yaml
@@ -18,7 +18,7 @@ image:
 sidecar:
   image:
     repository: 840364872350.dkr.ecr.us-west-2.amazonaws.com/aws-appmesh-envoy
-    tag: v1.22.2.1-prod
+    tag: v1.24.0.0-prod
     # sidecar.logLevel: Envoy log level can be info, warn, error or debug
   logLevel: info
   envoyAdminAccessPort: 9901
--- a/config/helm/appmesh-controller/test.yaml
+++ b/config/helm/appmesh-controller/test.yaml
@@ -7,6 +7,8 @@ region: ""
 accountId: ""
 preview: false
 clusterName: ""
+useAwsDualStackEndpoint: 0
+useAwsFIPSEndpoint: 0
 
 image:
   repository: 840364872350.dkr.ecr.us-west-2.amazonaws.com/amazon/appmesh-controller
--- a/config/helm/appmesh-controller/test.yaml
+++ b/config/helm/appmesh-controller/test.yaml
@@ -7,8 +7,8 @@ region: ""
 accountId: ""
 preview: false
 clusterName: ""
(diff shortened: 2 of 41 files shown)


pkg/aws/cloud.go
```diff
@@ -28,8 +28,9 @@ type Cloud interface {
 }
 
 // NewCloud constructs new Cloud implementation.
-func NewCloud(cfg CloudConfig, metricsRegisterer prometheus.Registerer) (Cloud, error) {
+func NewCloud(cfg CloudConfig, metricsRegisterer prometheus.Registerer, UseAwsDualStackEndpoint int, UseAwsFIPSEndpoint int) (Cloud, error) {
 	sess := session.Must(session.NewSession(aws.NewConfig()))
+	sessAPPMESH := session.Must(session.NewSession(aws.NewConfig()))
```
bendu, line 33: nit: can you change this to `sessAppMesh`

bendu: Also, can you add a comment explaining that due to the unavailability of dual stack endpoints (DNS ending in api.aws) that we have to have separate configs for App Mesh vs others

bendu (changes requested):

config/helm/appmesh-controller/values.yaml
```diff
@@ -8,6 +8,8 @@ accountId: ""
 preview: false
 enableBackendGroups: false
 clusterName: ""
+useAwsDualStackEndpoint: 0
+useAwsFIPSEndpoint: 0
```
BennettJames, line 12: These are essentially boolean flags, right? 

I believe the convention here is to keep them as boolean, then convert them to "0" / "1" specifically when needed if being converted to environment variables (same as you did in the prior FIPS patch).

It so happens that in this case, we need to situationally convert true/false values to 1/0. It's best to preserve the "true structure" of the boolean flags throughout the code so the meaning is not lost, then convert at the edges depending on what external systems expect.

pkg/aws/cloud.go
```diff
@@ -52,8 +53,19 @@ func NewCloud(cfg CloudConfig, metricsRegisterer prometheus.Registerer) (Cloud,
 		cfg.Region = region
 	}
 
-	awsCfg := aws.NewConfig().WithRegion(cfg.Region).WithSTSRegionalEndpoint(endpoints.RegionalSTSEndpoint)
+	awsCfgAPPMESH := &aws.Config{
```
BennettJames, line 56: Variables in Go should generally be [camel case](https://en.wikipedia.org/wiki/Camel_case); no reason for "appmesh" to be entirely caps here.

pkg/inject/config.go
```diff
@@ -55,6 +55,8 @@ const (
 	flagXRayImage            = "xray-image"
 
 	flagClusterName = "cluster-name"
+	flagUseAwsFipsEndpoint = "use-aws-fips-endpoint"
+	flagUseAwsDualStackEndpoint    = "use-aws-dual-stack-endpoint"
```
BennettJames, line 59: The formatting here doesn't quite look right - can you run this through gofmt? (e.g. `gofmt -w pkg/inject/config.go`). See [here](https://go.dev/blog/gofmt) for some details.

test/framework/framework.go
```diff
@@ -16,7 +16,6 @@ import (
 	"github.com/aws/aws-app-mesh-controller-for-k8s/test/framework/resource/virtualrouter"
 	"github.com/aws/aws-app-mesh-controller-for-k8s/test/framework/resource/virtualservice"
 	"github.com/aws/aws-app-mesh-controller-for-k8s/test/framework/utils"
-	. "github.com/onsi/gomega"
```
BennettJames, line 19: I think this line being removed is causing lots of compilation issues.

main.go
```diff
@@ -172,7 +172,7 @@ func main() {
 		setupLog.Error(err, "unable to start app mesh controller")
 		os.Exit(1)
 	}
-	cloud, err := aws.NewCloud(awsCloudConfig, metrics.Registry)
+	cloud, err := aws.NewCloud(awsCloudConfig, metrics.Registry, injectConfig.UseAwsDualStackEndpoint, injectConfig.UseAwsFIPSEndpoint)
```
BennettJames, line 175: Can the flags be placed on `aws.CloudConfig` instead of `inject.Config`? That would be closer to existing conventions, and eliminate the need to add two unstructured arguments here.

config/helm/appmesh-controller/test.yaml
```diff
@@ -7,6 +7,8 @@ region: ""
 accountId: ""
 preview: false
 clusterName: ""
+useAwsDualStackEndpoint: 0
+useAwsFIPSEndpoint: 0
```
BennettJames, line 11: Question - in practice, would it ever make sense to configure these differently than `sidecar.useDualStackEndpoint` and `sidecar.useFipsEndpoint`?

pkg/inject/config.go
joesbigidea: The makefile runs go fmt in the all and cotroller target. Are we not using the makefile for builds? If we're not we should figure out why.

config/helm/appmesh-controller/templates/deployment.yaml
```diff
@@ -79,6 +79,8 @@ spec:
         - --sds-uds-path={{ .Values.sds.udsPath }}
         - --enable-backend-groups={{ .Values.enableBackendGroups }}
         - --cluster-name={{ .Values.clusterName}}
+        - --use-aws-dual-stack-endpoint={{ .Values.useAwsDualStackEndpoint}}
+        - --use-aws-fips-endpoint={{ .Values.useAwsFIPSEndpoint}}
```
joesbigidea, line 83: We already have flags for use dual stack endpoint and use fips endpoint. Do we really need to add new flags for this?

pkg/inject/config.go
dileepng: I think make file runs go fmt. I have had merge conflicts when I am syncing my fork with main repo. may be that is the reason.

config/helm/appmesh-controller/templates/deployment.yaml
dileepng: The flags which we have already are being used only for sidecar. These are used for only AWS services (AppMesh,EKS,CloudMap). I think sidecar and AWS services are independent to each other. so I think we need to add new flags.

main.go
dileepng: I replaced flags from `inject.Config` to `aws.CloudConfig`.

EdwardXF (approved):

config/helm/appmesh-controller/templates/deployment.yaml
joesbigidea: I really wonder if a customer is every going to want to set these two sets of flags to different values. If a customer wants to use FIPS I'd expect they'd want to use it both in the sidecar and in the controller. It might be worth gathering some input on this, but it would be a real shame if we added two sets of flags that the customer always ends up changing together to get the desired result.

bendu: For FIPS, its straightforward to say we have just the one flag, but for dual stack its trickier. This is because we overlooked support for using dual stack endpoints in the controller, so we now have a breaking change where we are calling a new endpoint than the one we were calling before.

If we are ok with a slight breaking change for the dual stack endpoints (one might be able to argue this is really a bug fix since we are now doing what we say we are doing) then I think we should go ahead and use just the 1 set of flags.

bendu: But if we are not, then having the 2 sets of flags is fine

joesbigidea: I think I'd take the bugfix vs. break side of that argument, I really think we need to hold the line on making App Mesh harder to configure if at all possible.

dileepng: since other flags are already released, so we got to proceed with these flags for AWS api's.

joesbigidea (approved):

bendu (approved):

PR #667, opened 2022-12-22
Modifies integration tests to be runnable outside of AWS

BennettJames:
This patch reworks the test and CI system for integration tests to make the execution of the tests more platform agnostic. A system to manage and inject credentials is added that makes the tests runnable outside of an AWS context, and substantial adaptations are made to the test setup and runtime to make them run on a pure github action or linux environment.

Example integration test run: https://github.com/BennettJames/aws-app-mesh-controller-for-k8s/actions/runs/3756588966/jobs/6382826436
Example beta release run: https://github.com/BennettJames/aws-app-mesh-controller-for-k8s/actions/runs/3760250999/jobs/6391541156

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/actions/integration-test/action.yaml
+++ b/.github/actions/integration-test/action.yaml
@@ -1,0 +1,78 @@
+name: "Run Integration Test"
+description: "Runs integration tests in a kinD cluster on a github action"
+
+inputs:
+  aws_role:
+    description: "role to acquire from aws"
+    required: true
+  vpc_id:
+    description: "aws vpc id to use for the test"
+    required: true
+  account_id:
+    description: "aws account id to use"
+    required: true
+  cluster_name:
+    description: "name of the test cluster"
+    required: false
+    default: "test-cluster"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install Go 1.17
+      uses: actions/setup-go@v3
+      with:
+        go-version: '1.17.*'
+      id: go
+
+    - name: Setup Test Tools
+      shell: bash
+      run: |
+        export GOBIN=/usr/local/bin/
+
+        mkdir -vp ~/.docker/cli-plugins/
+        curl -sL -o ~/.docker/cli-plugins/docker-buildx "https://github.com/docker/buildx/releases/download/v0.3.0/buildx-v0.3.0.linux-amd64"
+        chmod a+x ~/.docker/cli-plugins/docker-buildx
+
+        curl -L -o kubebuilder.tar.gz "https://github.com/kubernetes-sigs/kubebuilder/releases/download/v2.3.1/kubebuilder_2.3.1_linux_amd64.tar.gz"
+        tar xzf kubebuilder.tar.gz
+        sudo mv "kubebuilder_2.3.1_linux_amd64" /usr/local/kubebuilder
+        export PATH=$PATH:/usr/local/kubebuilder/bin
+
+        curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3
+        chmod 700 get_helm.sh
+        ./get_helm.sh
+
+        ./scripts/install-controller-gen.sh
+        ./scripts/install-kubectl.sh
+        go install sigs.k8s.io/kustomize/kustomize/v4@latest
+
+        go mod download
+        go install github.com/onsi/ginkgo/ginkgo@v1.16.5
+
+    - name: Run Unit Tests
+      shell: bash
+      run: |
+        make test
+
+    - name: Configure AWS Credentials (build)
+      uses: aws-actions/configure-aws-credentials@v1-node16
+      with:
+        aws-region: us-west-2
+        role-to-assume: ${{ inputs.aws_role }}
+        role-session-name: IntegrationTest
+
+    - name: Setup Kind
+      uses: engineerd/setup-kind@v0.5.0
+      with:
+        version: "v0.17.0"
+        name: "${{ inputs.cluster_name }}"
+
+    - name: Run Integration Tests
+      shell: bash
+      env:
+        VPC_ID: "${{ inputs.vpc_id }}"
+        AWS_ACCOUNT_ID: "${{ inputs.account_id }}"
+        CLUSTER_NAME: "${{ inputs.cluster_name }}"
+      run: |
+        KUBECONFIG="${HOME}/.kube/config" ./scripts/test-with-kind.sh
--- a/.github/workflows/beta-release.yaml
+++ b/.github/workflows/beta-release.yaml
@@ -7,85 +7,57 @@ on:
         required: true
 
 permissions:
+  id-token: write
   contents: read
 
+env:
+  IMAGE_HOST: "${{ secrets.BETA_AWS_ACCOUNT }}.dkr.ecr.us-west-2.amazonaws.com"
+  IMAGE: "${{ secrets.BETA_AWS_ACCOUNT }}.dkr.ecr.us-west-2.amazonaws.com/amazon/appmesh-controller"
+  IMAGE_TAG: "${{ github.event.inputs.tag }}"
+  IMAGE_TAG_AMD: "${{ github.event.inputs.tag }}-linux_amd64"
+  IMAGE_TAG_ARM: "${{ github.event.inputs.tag }}-linux_arm64"
+
 jobs:
-  build:
-    name: integration-test
-    runs-on: [ self-hosted, aws-app-mesh-controller-for-k8s, X64 ]
+  integration-test:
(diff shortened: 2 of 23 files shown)


.github/workflows/beta-release.yaml
```diff
@@ -7,85 +7,65 @@ on:
...
 permissions:
+  id-token: write
   contents: read
 
+env:
+  IMAGE_HOST: "${{ secrets.CI_AWS_ACCOUNT }}.dkr.ecr.us-west-2.amazonaws.com"
+  IMAGE: "${{ secrets.CI_AWS_ACCOUNT }}.dkr.ecr.us-west-2.amazonaws.com/amazon/appmesh-controller"
+  IMAGE_TAG: "${{ github.event.inputs.tag }}"
+  IMAGE_TAG_AMD: "${{ github.event.inputs.tag }}-linux_amd64"
+  IMAGE_TAG_ARM: "${{ github.event.inputs.tag }}-linux_arm64"
```
BennettJames, line 18: Quick note on the variables here: before merging, I'd want to standardize on using just the beta account here. I'd like to get the patch mostly approved / reviewed, then create the appropriate roles in the beta account, then move over.

.github/actions/integration-test/action.yaml
```diff
@@ -0,0 +1,78 @@
+name: "Run Integration Test"
```
BennettJames, line 1: This is a more elaborate integration test setup that runs on plain ubuntu github actions. The setup behavior is much more complex than before (or rather - the complexity is explicit, rather than having been done out of sight on ec2 runners), and has been moved to a separate action so as to be callable from multiple locations.

.github/workflows/beta-release.yaml
```diff
@@ -7,85 +7,65 @@ on:
...
+          docker buildx build \
+            --platform linux/amd64 \
+            --build-arg GOPROXY="$GOPROXY" \
+            -t "${IMAGE}:${IMAGE_TAG}" \
+            . --push
+          docker buildx build \
+            --platform linux/arm64 \
+            --build-arg GOPROXY="$GOPROXY" \
+            -t "${IMAGE}:${IMAGE_TAG_ARM}" \
+            . --push
```
BennettJames, line 71: I had some earlier logic that did a proper multiarch deploy, but took it out to avoid the already rather broad scope. I can dial back the logic even further if desired to make this look more like the original action.

Dockerfile
```diff
@@ -1,21 +1,26 @@
...
 RUN go mod download
 
-COPY . ./
+COPY ./main.go ./ATTRIBUTION.txt ./
+COPY .git/ ./.git/
+COPY pkg/ ./pkg/
+COPY apis/ ./apis/
+COPY controllers/ ./controllers/
+COPY mocks/ ./mocks/
+COPY webhooks/ ./webhooks/
```
BennettJames, line 23: This wasn't strictly related, but I tweaked this dockerfile to be mildly more performant under some circumstances. Particularly: it uses the host build architecture for the building (which leads to fewer total actions in cross-building scenarios), and narrowed the scope of copied files to avoid spurious rebuilds.

test/framework/manifest/core_resources.go
```diff
@@ -17,6 +18,11 @@ type ManifestBuilder struct {
 
 	// required when serviceDiscoveryType == CloudMapServiceDiscovery
 	CloudMapNamespace string
+
+	// If set, will not enable ipv6 on any of the pods. This is needed to handle
+	// testing on github actions, where ipv6 is not supported -
+	// https://github.com/actions/runner-images/issues/668.
+	DisableIPv6 bool
```
BennettJames, line 25: This was a pretty annoying last minute surprise. I'm not happy with how that turned out - there's a lot of ugly logic in here to force ipv6 off in the integration tests to make it run on github actions. Otherwise, any test involving an actual envoy will hang during the proxyinit step displaying any error.

I'm tempted to add a separate flag to the controller itself to disable ipv6 on envoys. I'm worried that any additional work on the tests will result in future developers hitting the same problem, whereas a hard off switch for ipv6 in the controller itself would be more robust.

scripts/test-with-kind.sh
```diff
@@ -36,105 +43,135 @@ check_is_installed kubectl "You can install kubectl with the helper scripts/inst
...
+      --set sidecar.waitUntilProxyReady=true \
+      --set region="$AWS_REGION" \
+      --set accountId="$AWS_ACCOUNT_ID" \
+      --set "env.AWS_DEFAULT_REGION='$AWS_REGION'" \
+      --set "env.AWS_ACCESS_KEY_ID='$AWS_ACCESS_KEY_ID'" \
+      --set "env.AWS_SECRET_ACCESS_KEY='$AWS_SECRET_ACCESS_KEY'" \
+      --set "env.AWS_SESSION_TOKEN='$AWS_SESSION_TOKEN'" \
+      --set sidecar.envoyAwsAccessKeyId="$AWS_ACCESS_KEY_ID" \
+      --set sidecar.envoyAwsSecretAccessKey="$AWS_SECRET_ACCESS_KEY" \
+      --set sidecar.envoyAwsSessionToken="$AWS_SESSION_TOKEN"
```
BennettJames, line 154: Quick note on the permissions here:

This forwards the same set of credentials to two different locations: one, to the environment on the controller itself, and two, to the environment on the envoy containers. This could be finer grained - the envoy's need narrower permissions than the controller. This is needed as the runners lack the same AWS setup that let credentials be bootstrapped off the environment / metadata service.

I'm not thrilled by the system I added, and am very open to criticisms / suggestions here.

scripts/test-with-kind.sh
```diff
@@ -36,105 +43,135 @@ check_is_installed kubectl "You can install kubectl with the helper scripts/inst
...
+
+    check_deployment_rollout appmesh-controller appmesh-system
+    kubectl get pod -n appmesh-system
+
+    echo -n "running integration test type $__test ... "
+    ginkgo --flakeAttempts=2 -v -r $__test_dir -- \
```
BennettJames, line 160: I never like doing this, but I added an auto-retry to tests here w/ the `flakeAttempts` flag. At present, the test suite is just not super robust, and this increases the success rate considerably.

scripts/test-with-kind.sh
bendu: I think this looks reasonable to me for now. Since you have the credentials as separate here, we can always introduce another role later.

test/Readme.md
```diff
@@ -60,5 +60,13 @@ ginkgo -v -r test/integration/virtualnode/ -- --cluster-kubeconfig=/Users/xxxx/.
 In case of failures, refer to [Troubleshooting](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/docs/guide/troubleshooting.md) guide.   
 
 
+### integration test suite with kind
```
bendu, line 63: Thanks for adding instructions here. This will really help out the next person

test/framework/manifest/core_resources.go
bendu: I think both flag or hard off switch are acceptable. Since (I imagine) other customers have hit this ipv6 problem, what do you think about documenting this issue in a readme soemwhere?

.github/actions/integration-test/action.yaml
```diff
@@ -0,0 +1,78 @@
...
+description: "Runs integration tests in a kinD cluster on a github action"
+
+inputs:
+  aws_role:
+    description: "role to acquire from aws"
+    required: true
+  vpc_id:
+    description: "aws vpc id to use for the test"
+    required: true
+  account_id:
```
bendu, line 11: How are the inputs passed in? Will the account_id and role still be marked as secret?

bendu (approved):
I'm really happy to see this PR. Nice work!

amazing

.github/actions/integration-test/action.yaml
BennettJames: Yes - all inputs are elided in the same way (I think secrets are hidden based off of just pattern matching, so the origin doesn't really matter). You can see this in the [raw output](https://pipelines.actions.githubusercontent.com/serviceHosts/a48c4af4-3153-4cba-83b6-d8bb05fa32db/_apis/pipelines/1/runs/642/signedlogcontent/2?urlExpires=2022-12-22T23%3A45%3A52.5792756Z&urlSigningMethod=HMACV1&urlSignature=p2AbNwBv1CYj2VuKlFHxue9NdorPI7Sftx8CEkbh7fA%3D).

test/framework/manifest/core_resources.go
BennettJames: I think I may introduce a single flag w/ docs in a follow-up patch.

PR #668, opened 2022-12-23
Deploy a multiarch image

BennettJames:
Previously, the "main" image was always an amd64 image, and there was a secondary arm64 image that could be used with an explicit tag. This makes the amd64 image also have an explicit tag, and replaces the normal tagged image with a manifest file that will correctly route to the appropriate architecture.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/prod-release.yaml
+++ b/.github/workflows/prod-release.yaml
@@ -26,122 +26,101 @@ jobs:
       - name: Pull docker images
         run: |
           aws ecr get-login-password --region us-west-2 | docker login --username AWS --password-stdin ${{ secrets.BETA_AWS_ACCOUNT }}.dkr.ecr.us-west-2.amazonaws.com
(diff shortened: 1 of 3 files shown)


.github/workflows/prod-release.yaml
```diff
@@ -26,122 +26,101 @@ jobs:
...
+            region: "cn-north-1"
+            repo_host: "${{ secrets.PROD_BJS_AWS_ACCOUNT }}.dkr.ecr.cn-north-1.amazonaws.com.cn"
+            role: "${{ secrets.PROD_BJS_AWS_ROLE }}"
+
+      - name: Deploy Images to ZHY
+          uses: ./.github/actions/push-image
+          with:
+            source_image: "${{ secrets.BETA_AWS_ACCOUNT }}.dkr.ecr.us-west-2.amazonaws.com/amazon/appmesh-controller"
+            image_tag: "${{ github.event.inputs.tag }}"
+            region: "cn-north-1"
```
terranisu, line 124: I accidentally stumbled upon this PR. Shouldn't this value be `cn-northwest-1`? 🤔

bendu (changes requested):

Issue #619, opened 2022-08-05
Remove usage of busybox image when we want to inject Jaeger tracing


suniltheta:
**Is your feature request related to a problem?**
The controller injects busybox to add Jaeger config when user configures `--set tracing.provider=jaeger`. With newer App Mesh envoy version there is no need to inject this busy box image. The configuration file is generated with in the Envoy image itself.

**Describe the solution you'd like**
Get rid of injecting this image as it is redundant. We can also close https://github.com/aws/aws-app-mesh-controller-for-k8s/issues/405

Context: [#549 comment](https://github.com/aws/aws-app-mesh-controller-for-k8s/pull/549#issuecomment-1009514598)

PR #695, opened 2023-04-13
Removing usage of busybox image to add Jaeger

dileepng:
*Issue #, if available:*

*Description of changes:*
Removing the usage of busybox image when we want to use Jaeger tracing because with newer App Mesh envoy images the configuration will be generated by itself.

related issue: [more information ](https://github.com/aws/aws-app-mesh-controller-for-k8s/issues/619)


EdwardXF (approved):

suniltheta:
I see only change related to removing Jaeger config. Is there a change where we don't include busybox sidecar?

suniltheta (approved):

dileepng:
> I see only change related to removing Jaeger config. Is there a change where we don't include busybox sidecar?

busbox image is in initCointainer, which generates configuration for Jaeger tracer. so removed that initContainer.

BennettJames:
What versions of envoy is this change compatible with? We should probably announce a envoy version floor as part of the next release if this will be incompatible with older versions.

dileepng:
since from envoy version v1.16.1.0 this change is compatible with this configuration.

PR #706, opened 2023-05-26
Deploy a multiarch image

BennettJames:
*Issue #, if available:*

*Description of changes:*

Deploy a multiarch image

Previously, the "main" image was always an amd64 image, and there was
a secondary arm64 image that could be used with an explicit tag. This
makes the amd64 image also have an explicit tag, and replaces the normal
tagged image with a manifest file that will correctly route to the appropriate
architecture.


.github/workflows/prod-release.yaml
```diff
@@ -12,136 +12,92 @@ permissions:
...
-          role-to-assume: ${{ secrets.BETA_AWS_ROLE }}
-          role-session-name: ControllerProdRelease
-      - name: Pull docker images
-        run: |
-          aws ecr get-login-password --region us-west-2 | docker login --username AWS --password-stdin ${{ secrets.BETA_AWS_ACCOUNT }}.dkr.ecr.us-west-2.amazonaws.com
-          docker pull ${{ secrets.BETA_AWS_ACCOUNT }}.dkr.ecr.us-west-2.amazonaws.com/amazon/appmesh-controller:${{ github.event.inputs.tag }}
-          docker pull ${{ secrets.BETA_AWS_ACCOUNT }}.dkr.ecr.us-west-2.amazonaws.com/amazon/appmesh-controller:${{ github.event.inputs.tag }}-linux_arm64
-      - name: Configure AWS Credentials (prod)
-        uses: aws-actions/configure-aws-credentials@v1
+          src_host: "${{ secrets.BETA_AWS_ACCOUNT }}.dkr.ecr.us-west-2.amazonaws.com"
```
joesbigidea, line 26: You changed the beta-release workflow to use secrets.CI_AWS_ACCOUNT but here you're still using secrets.BETA_AWS_ACCOUNT. I could use some context to understand what's going on between these.

.github/workflows/beta-release.yaml
```diff
@@ -10,33 +10,11 @@ permissions:
...
-        uses: ./.github/actions/integration-test
-        with:
-          aws_role: "${{ secrets.BETA_TEST_AWS_ROLE }}"
-          vpc_id: "${{ secrets.INTEG_TEST_VPC }}"
-          account_id: "${{ secrets.BETA_AWS_ACCOUNT }}"
 
   push-images:
     name: Build And Push Images
     runs-on: ubuntu-22.04
-    needs: [ integration-test ]
```
joesbigidea, line 39: how are we enforcing integration-tests before beta push and eventually prod push?

BennettJames:
Example beta release: https://github.com/aws/aws-app-mesh-controller-for-k8s/actions/runs/5125981137/jobs/9220917101

Example prod release: https://github.com/aws/aws-app-mesh-controller-for-k8s/actions/runs/5125950425

.github/workflows/beta-release.yaml
BennettJames: Reverted. There are some process changes I want to make around testing for a release, but this isn't the right place to start them.

.github/workflows/prod-release.yaml
BennettJames: Standardized on BETA - these are the same values now; this patch just had it's roots back when CI was a distinct value that was used for beta image building and testing.

joesbigidea (approved):

Issue #710, opened 2023-06-23
Unable to create routes for VirtualRouter with multiple listener


brandocomando:
**Describe the bug**
When a VirtualRouter has more than one listener a port match is needed for creating a route, but match.port is not [supported](https://aws.github.io/aws-app-mesh-controller-for-k8s/reference/api_spec/#appmesh.k8s.aws/v1beta2.GRPCRouteMatch). 

**Steps to reproduce**
configure mesh "test-mesh"
configure virtual node "test-node"
create virtualRouter:
```
apiVersion: appmesh.k8s.aws/v1beta2
kind: VirtualRouter
metadata:
  name: test-router
spec:
  awsName: test-router
  listeners:
  - portMapping:
      port: 8080
      protocol: http
  - portMapping:
      port: 50002
      protocol: grpc
  meshRef:
    name: test-mesh
  routes:
  - httpRoute:
      action:
        weightedTargets:
        - port: 8080
          virtualNodeRef:
            name: test-node
          weight: 100
      match:
        port: 8080 #This can be added but does nothing
        prefix: /
    name: http
  - grpcRoute:
      action:
        weightedTargets:
        - port: 50002
          virtualNodeRef:
            name: test-node
          weight: 100
      match:
        port: 50002 #This can be added but does nothing
    name: grpc
```
**Expected outcome**
Virtual Router is created with 1 http route and 1 grpc route.

**Actual outcome**
```
[...]

PR #630, opened 2022-08-31
EKS multiple listener support

paulyeo21:
*Issue #, if available:*

https://github.com/aws/aws-app-mesh-roadmap/issues/120

*Description of changes:*

Support multiple listeners on EKS

*Testing:*

https://github.com/aws/aws-app-mesh-examples/pull/508

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #711, opened 2023-06-27
Preserve grpc port value from crd

BennettJames:
*Issue #, if available:*

*Description of changes:*

Fix behavior for #710: ensure port is carried over for grpc route targets in virtual router.

When multiple listener support was added in #630, all targets were updated to handle crd -> sdk conversion except grpc ports. This adds matching support for grpc.

*Testing:*

Re-created the issue; patched the controller; deployed it; confirmed it resolved it.

### DIFF
--- a/pkg/conversions/virtualrouter_types_conversion.go
+++ b/pkg/conversions/virtualrouter_types_conversion.go
@@ -333,8 +333,8 @@ func Convert_CRD_GRPCRouteMatch_To_SDK_GrpcRouteMatch(crdObj *appmesh.GRPCRouteM
 	sdkObj.Metadata = sdkMetadataList
 	sdkObj.MethodName = crdObj.MethodName
 	sdkObj.ServiceName = crdObj.ServiceName
+	sdkObj.Port = crdObj.Port
 	return nil
-
 }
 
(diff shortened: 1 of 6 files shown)


ysdongAmazon (approved):
Seems they forgot to include gRPC when enabled multiple listener last year. HTTP and TCP are good.

pkg/conversions/virtualrouter_types_conversion.go
```diff
@@ -333,8 +333,8 @@ func Convert_CRD_GRPCRouteMatch_To_SDK_GrpcRouteMatch(crdObj *appmesh.GRPCRouteM
 	sdkObj.Metadata = sdkMetadataList
 	sdkObj.MethodName = crdObj.MethodName
 	sdkObj.ServiceName = crdObj.ServiceName
+	sdkObj.Port = crdObj.Port
```
joesbigidea, line 336: Can we get tests on this behavior?

BennettJames:
> Seems they forgot to include gRPC when enabled multiple listener last year. HTTP and TCP are good.

Yup, added a few more details to the description to make it clearer.

pkg/conversions/virtualrouter_types_conversion.go
BennettJames: Very good point; updated.

Issue #712, opened 2023-07-19
GatewayRoute Admission Webhook not in keeping with AWS API/UI: match prefix does not need to end in '/'


conorevans:
**Describe the bug**

With the following YAML

```yaml
apiVersion: appmesh.k8s.aws/v1beta2
kind: GatewayRoute
metadata:
  name: conor-http
  namespace: mesh
spec:
  httpRoute:
    match:
      prefix: /conor
      port: 9088
    action:
      rewrite:
        hostname:
          defaultTargetHostname: DISABLED
        prefix:
          defaultPrefix: DISABLED
      target:
        virtualService:
          virtualServiceRef:
            name: conor
        port: 64001
```

I get the error

```zsh
✗ k apply -f gatewayroute.yaml
Error from server (Prefix to be matched on must start and end with '/'): error when creating "gatewayroute.yaml": admission webhook "vgatewayroute.appmesh.k8s.aws" denied the request: Prefix to be matched on must start and end with '/'
```

I do not get this in the UI:

![image](https://github.com/aws/aws-app-mesh-controller-for-k8s/assets/43791257/41fd7344-7ecd-498b-9385-a9290076591c)

Nor via the API, where the following Boto3 command:

```python
import boto3

client = boto3.client("appmesh")

response = client.create_gateway_route(
    gatewayRouteName="conor-http",
    meshName="mesh",
    spec={
        "httpRoute": {
            "action": {
                "rewrite": {
                    "hostname": {
                       "defaultTargetHostname": "DISABLED",
                    },
                    "prefix": {
                        "defaultPrefix": "DISABLED",
                    },
                },
                "target": {
                    "port": 64001,
                    "virtualService": {
                        "virtualServiceName": "conor.mesh.svc.cluster.local"
                    },
                },
            },
            "match": {
                "port": 9088,
                "prefix": "/conor",
            },
        },
    },
    virtualGatewayName="<redacted>",
)

print(response)
```

returns

```zsh
✗ python3 gatewayroute.py
[...]

PR #713, opened 2023-07-19
fix: httpRoute.match.prefix does not need to end in '/'

conorevans:
*Issue #, if available:* #712

*Description of changes:* Remove obligation for httpRoute.match.prefix to end in `/` - this does not match API behaviour.

### DIFF
--- a/webhooks/appmesh/gatewayroute_validator.go
+++ b/webhooks/appmesh/gatewayroute_validator.go
@@ -193,8 +193,8 @@ func validateHTTPRouteRewrite(rewrite *appmesh.HTTPGatewayRouteRewrite, match ap
 		return errors.New("Both prefix and path for rewrites cannot be specified. Only 1 allowed")
 	}
 	if rewrite.Prefix != nil {
-		if match.Prefix != nil && (!strings.HasSuffix(*match.Prefix, "/") || !strings.HasPrefix(*match.Prefix, "/")) {
-			return errors.New("Prefix to be matched on must start and end with '/'")
+		if match.Prefix != nil && !strings.HasPrefix(*match.Prefix, "/") {
+			return errors.New("Prefix to be matched on must start with '/'")
 		}
 		return validatePrefixRewrite(rewrite.Prefix)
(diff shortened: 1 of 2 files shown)


ysdongAmazon:
Checked the suffix pattern only exist under gateway route validation.
In our gateway route console and SDK, true, gateway route mentioned as prefix routing instead of path-based routing. Waiting for the team to confirm if it needs to be aligned with the SDK.

I suppose the original author was trying to ensure consistency in path routing: 
Prefix routing: consider two routes, one with prefix /product and another with prefix /product-special. If a request comes in with path /product-special/1234, both routes will match the request, because both prefixes are present in the path. This creates ambiguity about which route should process the request.

Path-based routing: enforced convention is applied and routes are defined with prefixes /product/ and /product-special/. With this convention, a request with path /product-special/1234 would only match the second route, eliminating the ambiguity.

ysdongAmazon:
Triggered unit test workflow first

conorevans:
> I suppose the original author was trying to ensure consistency in path routing:
> Prefix routing: consider two routes, one with prefix /product and another with prefix /product-special. If a request comes in with path /product-special/1234, both routes will match the request, because both prefixes are present in the path. This creates ambiguity about which route should process the request.

I can see that, but it breaks a very standard case, which is just a simple `/something` route. Even if I don't have `/something-else`, then I can't use `/something` because it isn't `/something/`. When you consider that someone in that case could simply use [HttpPathMatch](https://docs.aws.amazon.com/app-mesh/latest/APIReference/API_HttpPathMatch.html), I don't think this safeguard makes sense. Additionally, you could even use `priority` on the route even if you wanted to use Prefix, e.g. Prefix `/something-else` with priority 1 and Prefix `something` with priority 2, to ensure it's evaluated first. 

Will fix the unit test tomorrow!

Thanks for looking @ysdongAmazon 😄

conorevans:
@ysdongAmazon committed the test fix!

ysdongAmazon (approved):

PR #719, opened 2023-08-07
fxing TCP Route Match

dileepng:
*Description of changes:*

TCP Route Match is not required when we are creating a Tcp Route until we are configuring multiple listener. so making virtual route to be created with out TcpRoute Match.


EdwardXF (approved):

joesbigidea:
Could you add some tests around the optional route match? There are cases we didn't have before:
- existing route without match, add match
- existing route with match, remove match

PR #721, opened 2023-08-09
updating controller to include APPMESH_RESOURCE_ARN env variable

dileepng:
*Description of changes:*
Adding APPMESH_RESOURCE_ARN env variable to avoid discrepancies with the docs and leaving APPMESH_VIRTUAL_NODE_NAME because to make controller support older envoy versions.

### DIFF
--- a/pkg/inject/envoy_test.go
+++ b/pkg/inject/envoy_test.go
@@ -76,7 +76,7 @@ func Test_envoyMutator_mutate(t *testing.T) {
 			Namespace: "my-ns",
 			Name:      "my-pod",
 			Annotations: map[string]string{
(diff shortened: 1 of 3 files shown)


joesbigidea:
Could you provide some context on why you're making this change?

dileepng:
> Could you provide some context on why you're making this change?
our [docs](https://docs.aws.amazon.com/app-mesh/latest/userguide/envoy-config.html#envoy-required-config) are saying need to use APPMESH_RESOURCE_ARN from envoy version 1.15.0. Right now, latest controller version is v1.12.1 and it is injecting envoy v1.25, still controller still using APPMESH_VIRTUAL_NODE_NAME to set virtualnode/vg . so, customers are getting confused which one to be should be there, we got a ticket regarding this.

BennettJames (approved):

PR #723, opened 2023-08-17
update pdb api version

brandocomando:
*Issue #, if available:*

*Description of changes:*
PodDisruptionBudget.policy.v1beta1 has been deprecated in k8s 1.27
Update to policy/v1

### DIFF
--- a/config/helm/appmesh-controller/Chart.yaml
+++ b/config/helm/appmesh-controller/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: appmesh-controller
 description: App Mesh controller Helm chart for Kubernetes
-version: 1.12.3
-appVersion: 1.12.3
+version: 1.12.4
+appVersion: 1.12.4
 home: https://github.com/aws/eks-charts
 icon: https://raw.githubusercontent.com/aws/eks-charts/master/docs/logo/aws.png
 sources:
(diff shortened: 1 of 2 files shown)


config/helm/appmesh-controller/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: appmesh-controller
 description: App Mesh controller Helm chart for Kubernetes
-version: 1.12.3
-appVersion: 1.12.3
+version: 1.12.4
+appVersion: 1.12.4
```
BennettJames, line 5: Hey Brandon, we typically self-manage version incrementation. Given this is changing a version, we'd likely cut a new minor version for it.

config/helm/appmesh-controller/templates/pdb.yaml
```diff
@@ -1,7 +1,7 @@
 {{- if .Values.podDisruptionBudget }}
 {{- if gt (int .Values.replicaCount) 1 }}
 kind: PodDisruptionBudget
-apiVersion: policy/v1beta1
+apiVersion: policy/v1
```
BennettJames, line 4: Just a note: I'm ok taking this once the version change for helm is removed, but I'll add that we'll probably need to do some inter-version testing for this before doing a full release. These sorts of inter-version incompatibility issues often create problems, and we'd need to be confident we have complete guidance for compatibility before doing a general release.

PR #725, opened 2023-08-28
Add length validation to route name

geezyx:
*Description of changes:*

According to the API reference, the route name must be at least 1 character and at most 255 characters. https://docs.aws.amazon.com/app-mesh/latest/APIReference/API_RouteData.html

Currently an empty string is valid input for the CRD, which results in an API error during reconcile. This can be caught early with a validation rule.


ysdongAmazon:
Thank you, Mike for pointing out. Executed unit tests

Issue #724, opened 2023-08-21
can't inject env var with a comma in it's value


Omar-Bishtawi:
**Describe the bug**
if you try to inject the following annotation to any deployment 
```
appmesh.k8s.aws/sidecarEnv: '[{"sample_key_1":"value","sample_key_2":"value"}]'
```

you will get the follwing error
```
malformed annotation appmesh.k8s.aws/sidecarEnv , expected format: EnvVariableKey=EnvVariableValue
```
**Steps to reproduce**
attach any json value with multiple keys to sidecar env var . 

**Expected outcome**
the env variable to have a valid json as it's value

**Additional Context**:
I think this behavior is expected since in [here](https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/c27086da1ddf629305f769bf5aaacb390bf41985/pkg/inject/envoy.go#L238)the method only split by `,` and we can't escape this charachter in any way.

Omar-Bishtawi:
are you open to the idea of supporting such  feature (allow escaping of the `,`  character if it wasn't ment to be used as a separator)  ?

BennettJames:
Hey Omar,

We're not necessarily against improving the behavior here, but it's tricky to do in a fairly general way that's reasonably backwards compatible. We'd likely need to adapt a system that's capable of parsing the string such that `,`, `"`,  and `=` are special characters that control handling, and allow escaping of them. Maybe it'd be ok breaking some edge cases like if a current user is setting variables like `A=ab"c\,B=efg`, but I'm always pretty hesitant to make a change that can have unexpected consequences.

Perhaps the best answer is to add a new variable that has a more robust parsing system around it - maybe `sidecarEnvJson` - and let the user specify a json map of strings-to-values.

Omar-Bishtawi:
@BennettJames  the idea of introducing a new env variable with robust parsing system is good one and resolve any backward compatibility issues that might arise of modifying the behavior of the old `sidecarEnv`. 

also using JSON as a way to map env vars with there values is more robust and easier to parse in a consistent expected way.

PR #726, opened 2023-08-30
adding sidecarEnvJson annotation

dileepng:
*Issue #, if available:*
https://github.com/aws/aws-app-mesh-controller-for-k8s/issues/724
*Description of changes:*
Adding Feature appmesh.k8s.aws/sidecarEnvJson that accepts json list as env for envoy.

### DIFF
--- a/pkg/inject/envoy.go
+++ b/pkg/inject/envoy.go
@@ -2,6 +2,7 @@ package inject
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/util/json"
 	"strconv"
 	"strings"
 
@@ -90,6 +91,15 @@ func (m *envoyMutator) mutate(pod *corev1.Pod) error {
 	customEnv, err := m.getCustomEnv(pod)
 	if err != nil {
 		return err
+	}
+
+	customEnvJson, err := m.getCustomEnvJson(pod)
+	if err != nil {
+		return err
+	}
+
+	for k, v := range customEnvJson {
+		customEnv[k] = v
 	}
 
 	container, err := buildEnvoySidecar(variables, customEnv)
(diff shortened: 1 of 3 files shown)


pkg/inject/envoy.go
```diff
@@ -249,6 +259,45 @@ func (m *envoyMutator) getCustomEnv(pod *corev1.Pod) (map[string]string, error)
 	return customEnv, nil
 }
 
+type customEnvJsonType []map[string]string
+
+func (c *customEnvJsonType) UnmarshalJSON(data []byte) error {
```
ysdongAmazon, line 264: Need we make a valid check for bytes before unmarshal if the input is empty or nil?

dileepng: It will throw malformed error. so, I didn't make a check for that condition.

ysdongAmazon: Thanks. Good to know

pkg/inject/envoy.go
```diff
@@ -249,6 +259,45 @@ func (m *envoyMutator) getCustomEnv(pod *corev1.Pod) (map[string]string, error)
 	return customEnv, nil
 }
 
+type customEnvJsonType []map[string]string
```
joesbigidea, line 262: why are you not just using a map[string]string?

pkg/inject/envoy.go
```diff
@@ -249,6 +259,41 @@ func (m *envoyMutator) getCustomEnv(pod *corev1.Pod) (map[string]string, error)
 	return customEnv, nil
 }
 
+type customEnvJsonType map[string]string
```
BennettJames, line 262: Does this custom type still need to exist, or can `getCustomEnvJson` just deserialize to `map[string]string` directly?

BennettJames: The only place I see this type used is in the `getCustomEnvJson` function, which immediately unwraps it to a `map[string]string`.

Generally, you only need to some serialization-time changes between the representative data type and the target data type (for instance, when converting strings to dates). I think that might have been true of the first draft of this where the input data was a list of strings, but now that it's just a map of string->string it's the same as what the target type is. Just passing `map[string]string` to unmarshal should accomplish this fine without the indirection.

BennettJames (approved):

Issue #767, opened 2024-03-21
Cannot update VirtualRouter listener protocol with defined routes


dhild:
**Describe the bug**
If I replace the listener protocol and a matching route's protocol in a `VirtualRouter` at the same time, the change is accepted by the API server but fails to reconcile.

**Steps to reproduce**

1. Create a `VirtualRouter` with a listener and a matching route. Wait for the status to show that the `VirtualRouter` is successfully reconciled.
```
kind: VirtualRouter
apiVersion: appmesh.k8s.aws/v1beta2
metadata:
  name: example-router
spec:
  listeners:
  - portMapping:
      protocol: http
      port: 8080
  routes:
  - name: web
    httpRoute:
      match:
        prefix: /
        port: 8080
      action:
        weightedTargets:
        - virtualNodeRef:
            name: example-node
          port: 8080
          weight: 1
```
2. Change the listener protocol, and update the protocol used for the route to match. This change will be accepted by the API server.
```
kind: VirtualRouter
apiVersion: appmesh.k8s.aws/v1beta2
metadata:
  name: example-router
spec:
  listeners:
  - portMapping:
      protocol: http2
      port: 8080
  routes:
  - name: web
    http2Route:
      match:
        prefix: /
        port: 8080
      action:
        weightedTargets:
        - virtualNodeRef:
            name: example-node
          port: 8080
          weight: 1
```
3. Warning events can be observed on the `VirtualRouter`, and the status never transitions to observe the latest generation.
```
  Warning  ReconcileError  5s   VirtualRouter  BadRequestException: 1 Virtual Router listener(s) cannot be removed because they are matched by existing Routes. Listing up to 5 PortMappings: [(Port: 8080, Protocol: HTTP)]
```

**Workaround**
Manually sequence the updates, by first removing the route before changing the listener protocol. Once this has reconciled, apply the desired final spec.
```
kind: VirtualRouter
apiVersion: appmesh.k8s.aws/v1beta2
metadata:
  name: example-router
spec:
  listeners:
  - portMapping:
      protocol: http
      port: 8080
  routes: []
```

**Expected outcome**
The controller should perform the API calls in the same sequence as the workaround, leaving the AWS resources in the desired state.

**Environment**

* App Mesh controller version: v1.12.1
* Envoy version: v1.25.4.0-prod
* Are you using any integrations? X-ray, Jaeger etc. If so versions? N/A
* Kubernetes version v1.27.10
[...]

PR #768, opened 2024-03-21
Fix VirtualRouter route deletion sequence

dhild:
This addresses a few rough edges with updating a `VirtualRouter`, by removing routes which will be deleted before updating the listeners. After this change, it is possible to update the listener protocol on the CRD in one step. It is also possible to remove the listener at the same time as the routes which use that listener.

*Issue #, if available:* #767

*Description of changes:*

Adds a step to the `VirtualRouter` reconciliation sequence which identifies the routes to remove and deletes them before the corresponding listeners are updated. This allows a greater range of valid API changes to be successfully reconciled.

### DIFF
--- a/pkg/virtualrouter/routes_manager.go
+++ b/pkg/virtualrouter/routes_manager.go
@@ -295,13 +295,13 @@ func taintedSDKRouteRefs(routes []appmesh.Route, sdkVR *appmeshsdk.VirtualRouter
 
 	for _, name := range matchedNameSet.List() {
 		route := routeByName[name]
-		if route.TCPRoute != nil && route.TCPRoute.Match != nil && sdkListenerByPort[aws.Int64Value(route.TCPRoute.Match.Port)] != appmesh.PortProtocolTCP {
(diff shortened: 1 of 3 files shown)


pkg/virtualrouter/routes_manager.go
```diff
@@ -249,6 +273,47 @@ func matchRoutesAgainstSDKRouteRefs(routes []appmesh.Route, sdkRouteRefs []*appm
...
+	}
+
+	unmatchedSDKRouteRefs := make([]*appmeshsdk.RouteRef, 0, len(unmatchedSDKRouteRefNameSet))
+	for _, name := range unmatchedSDKRouteRefNameSet.List() {
+		unmatchedSDKRouteRefs = append(unmatchedSDKRouteRefs, sdkRouteRefByName[name])
+	}
+
+	return unmatchedSDKRouteRefs
+}
+
```
BennettJames, line 316: Could we get unit tests covering `remove` and this function?

dhild: Sure, I'll get some test coverage written up for those.

BennettJames (approved):

BennettJames:
Merged.

It might be a bit before we release this as the default, but we can try to put up a version you can use in the meantime with a helm override if that'd be useful.

dhild:
Great, thanks! I would appreciate having an official image to run these changes in test environments, if it’s not too much trouble.

PR #776, opened 2024-05-21
adding feature to set TLS configuration

kangsheng89:
*Issue #, if available:*

*Description of changes:*
This is enable user to configure TLS setting for the controller manager
- setting for minimum TLS version
- setting for preferred TLS cipher suites

### DIFF
--- a/main.go
+++ b/main.go
@@ -19,10 +19,11 @@ package main
 import (
 	"context"
 	"crypto/tls"
-	"github.com/aws/aws-sdk-go/service/eks"
 	"os"
 	"strconv"
 	"time"
+
+	"github.com/aws/aws-sdk-go/service/eks"
 
 	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/aws/throttle"
 	"github.com/aws/aws-app-mesh-controller-for-k8s/pkg/cloudmap"
@@ -332,6 +333,7 @@ func tlsConfigSetting(cfg *tls.Config, tlsOpt tlsConfig) {
 	tlsVersion, err := k8sapiflag.TLSVersion(tlsOpt.minVersion)
 	if err != nil {
 		setupLog.Error(err, "TLS version invalid")
+		os.Exit(1)
 	}
 	cfg.MinVersion = tlsVersion
 
@@ -339,6 +341,7 @@ func tlsConfigSetting(cfg *tls.Config, tlsOpt tlsConfig) {
 	cipherSuiteIDs, err := k8sapiflag.TLSCipherSuites(tlsOpt.cipherSuites)
 	if err != nil {
 		setupLog.Error(err, "Failed to convert TLS cipher suite name to ID")
+		os.Exit(1)
 	}
 	cfg.CipherSuites = cipherSuiteIDs
 }
--- a/main.go
+++ b/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
 	"github.com/aws/aws-sdk-go/service/eks"
 	"os"
 	"strconv"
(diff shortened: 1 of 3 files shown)


kangsheng89:
@ysdongAmazon could help to review this?

sshver (approved):

kangsheng89:
@sshver could help to approve and trigger the workflow?

kangsheng89:
@sshver once this done, do i require to update helm chart at https://github.com/aws/eks-charts/tree/master/stable/appmesh-controller?

main.go
```diff
@@ -308,3 +323,22 @@ func main() {
...
 }
+
+// This function get the option from command argument (tlsConfig), check the validity through k8sapiflag
+// and set the config for webhook server.
+// refer to https://pkg.go.dev/k8s.io/component-base/cli/flag
+func tlsConfigSetting(cfg *tls.Config, tlsOpt tlsConfig) {
+	// TLSVersion helper function returns the TLS Version ID for the version name passed.
+	tlsVersion, err := k8sapiflag.TLSVersion(tlsOpt.minVersion)
+	if err != nil {
+		setupLog.Error(err, "TLS version invalid")
```
joesbigidea, line 334: I think we should be exiting on failure to honor this setting. The same applies to the cipher suites.

main.go
```diff
@@ -105,6 +113,8 @@ func main() {
 		"The page size limiting the number of response for list operation to API Server")
 	fs.IntVar(&healthProbePort, flagHealthProbePort, defaultHealthProbePort,
 		"The port the health probes binds to.")
+	fs.StringVar(&tlsOpt.minVersion, "tls-min-version", "VersionTLS12", "Minimum TLS version supported. Value must match version names from https://golang.org/pkg/crypto/tls/#pkg-constants.")
```
joesbigidea, line 116: It would be best to leave the default behavior a no-op when tls-min-version is not specified to avoid any impact to existing deployments.

main.go
```diff
@@ -308,3 +323,22 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+// This function get the option from command argument (tlsConfig), check the validity through k8sapiflag
+// and set the config for webhook server.
+// refer to https://pkg.go.dev/k8s.io/component-base/cli/flag
+func tlsConfigSetting(cfg *tls.Config, tlsOpt tlsConfig) {
```
joesbigidea, line 330: This probably warrants a unit test.

joesbigidea (changes requested):
Thanks for submitting this! I have a couple of small changes that would be good to include.

main.go
joesbigidea: Also it would be good to handle the interpreting of the tls options in main up front, and just pass the values into this function. That would keep any argument handling logic in a single place, and make sure it happens right away on startup similar to other argument parsing failures.

main.go
kangsheng89: If we set to "nil", it will fall back to default setting from k8sapiflag DefaultTLSVersion() https://github.com/kubernetes/component-base/blob/v0.26.2/cli/flag/ciphersuites_flag.go#L142 which is the same as tls1.2.

main.go
kangsheng89: Thank for highlighting on exiting on failure. For second point, my opinion is that when NewManager is called, it will callback on setting the TLS config, if we exit with failure, the controller will not able to start anyway. 
May b u could show me how u organize the argument handling logic and i could refer to it. Hehe..  I am not the expert in this domain and need your help to guide me.

main.go
kangsheng89: @joesbigidea i dint see any unit-test current for the main.go from makefile refer to https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/Makefile#L35. Could u help me on this?

joesbigidea:
@kangsheng89 Since we want to get this change out quickly @sshver made a few small structural changes to this in a separate PR. That PR has been manually tested and we'll go ahead and get that one merged and released. Thank you for your contribution.

kangsheng89:
Thank you! @joesbigidea and @sshver

